PR #9, opened 2019-08-08
D8 nid 225

omahm:


nidirect_money_advice_articles/src/ArticleProcessors.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+class ArticleProcessors {
+
+  /**
+   * Removes summary text from body value.
+   */
+  public static function body($value) {
+    $pattern = '/<p><strong>.*?<\/strong><\/p>(.*?)$/s';
+    if (preg_match($pattern, $value, $matches)) {
+      return $matches[1];
```
johangant: Finding myself pause on this a bit. Just so I'm clear... this does appear to remove the summary text, but it does this by group matching the text after the initial <p><strong></strong></p> (assumed to contain the summary) and only returning the matched content?

omahm: Yeah, pretty much a copy from the original tamper code. Ignores the summary part of the text and returns everything after.

PR #11, opened 2019-08-09
D8 nid 87

DuttonMa:
- Add 'needs audit' view to dashboard
- Implement audit functionality driven by dates to replace old implementation that used Flags, Rules and Rules Scheduler.
- Implement a custom 'audit content' permission which controls access to auditing
- Implement NIDirect admin screen under 'Configuration' that allows audit strings to be maintained

### DIFF
--- a/nidirect_workflow/nidirect_workflow.module
+++ b/nidirect_workflow/nidirect_workflow.module
@@ -46,13 +46,13 @@ function nidirect_workflow_entity_presave(EntityInterface $entity) {
 function nidirect_workflow_form_alter(&$form, FormStateInterface $form_state, $form_id) {
 
   // Manage access to 'next audit date' field.
-  if (($form_id == 'node_article_form')
-    || ($form_id == 'node_article_edit_form')
-    || ($form_id == 'node_contact_form')
-    || ($form_id == 'node_contact_edit_form')
-    || ($form_id == 'node_page_form')
-    || ($form_id == 'node_page_edit_form')
-  ) {
+  if (in_array($form_id,[
+    'node_article_form',
+    'node_article_edit_form',
+    'node_contact_form',
+    'node_contact_edit_form',
+    'node_page_form',
+    'node_page_edit_form'])) {
     $account = \Drupal::entityTypeManager()->getStorage('user')->load(\Drupal::currentUser()->id());
     // Check user access level.
     if ($account->hasRole('administrator')) {
--- a/nidirect_workflow/nidirect_workflow.module
+++ b/nidirect_workflow/nidirect_workflow.module
@@ -39,8 +39,6 @@ function nidirect_workflow_entity_presave(EntityInterface $entity) {
             $entity->set('field_next_audit_due', date('Y-m-d', strtotime("+6 months")));
           }
         }
-        // - Need to add the 'audit this content' link to the node view page
-        // - Need to restrict access to the field in node edit.
         break;
 
       case 'health_condition':
@@ -49,7 +47,7 @@ function nidirect_workflow_entity_presave(EntityInterface $entity) {
           $review_date = $entity->get('field_next_review_date')->value;
           if (!empty($review_date)) {
             // Set 'audit due' date to be the same.
-            $next_audit_date = $entity->get('field_next_audit_due')->value;
+            $entity->set('field_next_audit_due', $review_date);
           }
         }
     }
@@ -62,7 +60,15 @@ function nidirect_workflow_entity_presave(EntityInterface $entity) {
 function nidirect_workflow_form_alter(&$form, FormStateInterface $form_state, $form_id) {
 
   // Disable the 'next audit date' for all but admins.
-  if (($form_id == 'node_article_form') || ($form_id == 'node_article_edit_form')) {
+  if (($form_id == 'node_article_form')
+    || ($form_id == 'node_article_edit_form')
+    || ($form_id == 'node_contact_form')
+    || ($form_id == 'node_contact_edit_form')
+    || ($form_id == 'node_page_form')
+    || ($form_id == 'node_page_edit_form')
+    || ($form_id == 'node_health_condition_form')
+    || ($form_id == 'node_health_condition_edit_form')
+  ) {
     $account = User::load(\Drupal::currentUser()->id());
     if (!$account->hasRole('administrator')) {
       $form['field_next_audit_due']['#disabled'] = 'TRUE';
--- a/nidirect_workflow/nidirect_workflow.module
+++ b/nidirect_workflow/nidirect_workflow.module
@@ -59,7 +59,7 @@ function nidirect_workflow_entity_presave(EntityInterface $entity) {
  */
 function nidirect_workflow_form_alter(&$form, FormStateInterface $form_state, $form_id) {
 
-  // Disable the 'next audit date' for all but admins.
+  // Manage access to 'next audit date' field.
   if (($form_id == 'node_article_form')
     || ($form_id == 'node_article_edit_form')
     || ($form_id == 'node_contact_form')
@@ -70,8 +70,21 @@ function nidirect_workflow_form_alter(&$form, FormStateInterface $form_state, $f
     || ($form_id == 'node_health_condition_edit_form')
   ) {
     $account = User::load(\Drupal::currentUser()->id());
-    if (!$account->hasRole('administrator')) {
-      $form['field_next_audit_due']['#disabled'] = 'TRUE';
+    // Check user access level.
+    if ($account->hasRole('administrator')) {
+      // Admin can see audit date and change it.
+      $form['field_next_audit_due']['#access'] = TRUE;
+      $form['field_next_audit_due']['#disabled'] = FALSE;
+    }
+    elseif ($account->hasPermission('audit content')) {
+      // If user has permission to audit content but is not an admin
+      // then they can see the audit date but not change it.
+      $form['field_next_audit_due']['#access'] = TRUE;
+      $form['field_next_audit_due']['#disabled'] = TRUE;
+    }
+    else {
+      // Hide the audit date for anyone else.
+      $form['field_next_audit_due']['#access'] = FALSE;
     }
   }
 }
--- a/nidirect_workflow/nidirect_workflow.module
+++ b/nidirect_workflow/nidirect_workflow.module
@@ -10,10 +10,6 @@ use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
 
 /**
- * @file
- */
-
-/**
  * Implements hook_page_attachments().
  */
 function nidirect_workflow_page_attachments(array &$attachments) {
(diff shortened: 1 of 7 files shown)


nidirect_workflow/nidirect_workflow.module
```diff
@@ -11,3 +20,59 @@ function nidirect_workflow_page_attachments(array &$attachments) {
   // Attach extra custom css for admin menu.
   $attachments['#attached']['library'][] = 'nidirect_workflow/admin.css';
 }
+
+/**
+ * Implements hook_entity_presave().
+ */
+function nidirect_workflow_entity_presave(EntityInterface $entity) {
```
johangant: A suggestion: consider the use of an early return pattern to limit the amount of depth/indentation in the code here.

So instead of:

```
function thingy() {
  if ($thing == 'a-certain-value') {
    // more checks/things to do...
    return $value;
  }
}
```

try:

```
function thingy() {
  if ($thing != 'a-specific-value') {
    return;
  }

  // Anything from here is useful code/contextually relevant. Otherwise, we bail early. Bonus points for not needing an extra level of indentation, keeping it more readable.

  // Return a thing.
  return $thing;
}
```

nidirect_workflow/nidirect_workflow.module
```diff
@@ -11,3 +20,59 @@ function nidirect_workflow_page_attachments(array &$attachments) {
...
+  }
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function nidirect_workflow_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+
+  // Manage access to 'next audit date' field.
+  if (($form_id == 'node_article_form')
```
johangant: I'd pop these into an array and run a single: `if (in_array($values, $form_id) { .. }` or alternatively `if (preg_match('/pattern/', $form_id) { .. }` for brevity.

Caveat: I can never remember the order of parameters in `in_array()` :D

nidirect_workflow/nidirect_workflow.module
```diff
@@ -11,3 +20,59 @@ function nidirect_workflow_page_attachments(array &$attachments) {
...
+    || ($form_id == 'node_article_edit_form')
+    || ($form_id == 'node_contact_form')
+    || ($form_id == 'node_contact_edit_form')
+    || ($form_id == 'node_page_form')
+    || ($form_id == 'node_page_edit_form')
+  ) {
+    $account = User::load(\Drupal::currentUser()->id());
```
johangant: Better to load the user via the entity type manager service? That way it's consistent with other entity types.

nidirect_workflow/src/Controller/AuditController.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
```
johangant: No @file block comment. That'll fail PHPCS ;)

nidirect_workflow/src/Controller/AuditController.php
```diff
@@ -0,0 +1,110 @@
...
+   */
+  public function contentAudit($nid) {
+    $msg = 'Invalid node id';
+    if (!empty($nid)) {
+      $node = $this->entityTypeManager()->getStorage('node')->load($nid);
+      if ($node) {
+        $audit_button_text = $this->config('nidirect_workflow.auditsettings')->get('audit_button_text');
+        $audit_confirmation_text = $this->config('nidirect_workflow.auditsettings')->get('audit_confirmation_text');
+        $msg = $this->t($audit_confirmation_text);
+        $msg .= "<div><a href='/nidirect_workflow/confirm_audit/$nid?destination=/node/$nid'>" . $this->t($audit_button_text) . "</a></div>";
```
johangant: HTML in the controller doesn't feel right to me. I'd consider using a Link/Url building service here and move the divs etc out into a theme function/template.

nidirect_workflow/src/Controller/AuditController.php
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Content Audit.
+   *
+   * @return string
+   *   Return confirmation string.
+   */
+  public function contentAudit($nid) {
+    $msg = 'Invalid node id';
```
johangant: Wrap in `$this->t()` for translation. Does node id need to be node ID? Have a look in core and see what people have used there.

nidirect_workflow/src/Controller/AuditController.php
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Confirm Audit.
+   *
+   * @return string
+   *   Return confirmation string.
+   */
+  public function confirmAudit($nid) {
+    // Bump up the 'next audit due' date and log it.
+    $node = $this->entityTypeManager()->getStorage('node')->load($nid);
+    if ($node) {
```
johangant: Isn't this always going to be a node? Except when the entity type manager can't load something in which case it'll be null?

Might be semantically clearer to use `if (empty($node)) { ... }`?

nidirect_workflow/src/Form/AuditSettingsForm.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
```
johangant: Does PHPCS pass without a `@file` block comment?

nidirect_workflow/nidirect_workflow.module
DuttonMa: Hmm, yes, interesting (and debatable).

I think that I'm still stuck in the old school idea of "only one return in a function" but some Googling suggests that this is a bit outdated now. I must say though that I am still reluctant to have many return statements in a long and complex function (perhaps there shouldn't be any long and complex functions anywhere, that's a separate discussion?!).

I do agree that this code looks a bit ugly so will go with your suggestion here.

nidirect_workflow/nidirect_workflow.module
DuttonMa: Personally I prefer it as is - I think it is more readable !

nidirect_workflow/nidirect_workflow.module
DuttonMa: Yes, very good point - I'm still getting used to this EntityTypeManager, I need to use it more!

nidirect_workflow/src/Controller/AuditController.php
DuttonMa: Will add a comment after namespace as discussed

nidirect_workflow/src/Controller/AuditController.php
DuttonMa: Yes, good spot

nidirect_workflow/src/Form/AuditSettingsForm.php
DuttonMa: Yes it does, but will add a comment as discussed

nidirect_workflow/src/Controller/AuditController.php
DuttonMa: Matter of opinion really, I prefer:

  if ($node)

to:

  if (!empty($node))

as I think it scans better.
And of course it could also be:

  if (isset($node))

nidirect_workflow/src/Controller/AuditController.php
DuttonMa: Hmm, well, I kind of see what you mean but I'm not sure that I agree.

We are in a controller here and we are returning markup, so it is bound to be HTML isn't it?

If you have a look at DbUpdateController.php in core you'll see that there is plenty of HTML in there.

nidirect_workflow/nidirect_workflow.module
DuttonMa: Have had another look at this and actually think I'll leave it as is, early returns still don't sit well with me !

nidirect_workflow/nidirect_workflow.module
omahm: I'd be more of a fan of in_array()
Happy medium could be:

  if (in_array($form_id, [
    'node_article_edit_form',
    'node_contact_form',
    'node_contact_edit_form'])) {

  }

nidirect_workflow/src/Controller/AuditController.php
omahm: Also, add the $nid to the message.

DuttonMa:
Thanks for the comments Johan, but I think that we should think about the actual purpose of a code review.

In my mind a code review should definitely pick up things like missing translations as there is no debate about that.

I think, though, that to discuss whether:

    if ($node)

should be changed to:

    if (!empty($node))

is just a question of style. Neither way is right or wrong so debating it is just a waste of time!

We each have our own coding styles and I don't think that we should try to change that.

nidirect_workflow/src/Controller/AuditController.php
omahm: Yeah, I'd use https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Link.php/class/Link/8.2.x
as it returns a render array that if it ever needs overridden it can easily be done.

nidirect_workflow/src/Controller/AuditController.php
omahm: It'll return the entity or null so I think if($node) reads easier.

johangant:
Cheers for reviewing @DuttonMa. Not a huge problem about differences in style, so I think we can park those. I'm sure they can be adjusted in future if they ever need to.

I _think_ that just leaves some things around Link class usage. I don't think there's a service for that so you'll likely only need to reference that static class to generate the render arrays you need.

omahm:
> I _think_ that just leaves some things around Link class usage. I don't think there's a service for that so you'll likely only need to reference that static class to generate the render arrays you need.

Just create the second parameter for the link using the URL object which will allow you set set options such as classes etc.

nidirect_workflow/src/Controller/AuditController.php
DuttonMa: Ahh, yes, of course. Apologies Johan I misunderstood you, I thought that you were suggesting that _I_ should write a Link/Url building service! (Hence my overkill comment.)

I do agree with you both that I should use the Drupal Link object, I'll go ahead and do that.

nidirect_workflow/nidirect_workflow.module
DuttonMa: OK, will go with that

nidirect_workflow/src/Controller/AuditController.php
```diff
@@ -101,6 +116,7 @@ public function confirmAudit($nid) {
       $message = "nid " . $nid . " " . $this->t("has been audited by") . " ";
       $message .= $this->account->getAccountName() . " (uid " . $this->account->id() . ")";
       $this->logger->notice($message);
+      \Drupal::messenger()->addMessage(t('Content has been marked as audited'));
```
omahm: The messenger service should be injected into the controller as a dependency, also of note is: https://www.drupal.org/node/2928994

DuttonMa: Wow, you're on the ball this morning - I hadn't even requested another review yet !

Good point, but looking through core it seems that I don't have to inject the messenger as it is already available in ControllerBase, so have just changed it to:

$this->messenger()->addMessage

omahm: haha, that's true, good catch.

PR #12, opened 2019-08-09
Add a search breadcrumb handler class

johangant:
Search results pages do not show a breadcrumb. Rather than preprocess the heck out the template for that page (as per D7) we can introduce a new breadcrumb handler class that responds to the route name and can, in this case, set an empty array of links which in turn means nothing is rendered.

In future, if we need to adjust other views displays - we only need to put some more conditional checks on the route and can adapt what is displayed in each instance.

### DIFF
--- a/nidirect_breadcrumbs/src/SearchBreadcrumb.php
+++ b/nidirect_breadcrumbs/src/SearchBreadcrumb.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Drupal\nidirect_breadcrumbs;
+
+/**
+ * @file
+ * Generates the breadcrumb trail for search page(s)
+ */
+
+use Drupal\Core\Breadcrumb\Breadcrumb;
+use Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Link;
+use Drupal\Core\Url;
+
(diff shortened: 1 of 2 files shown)


nidirect_breadcrumbs/src/SearchBreadcrumb.php
```diff
@@ -0,0 +1,43 @@
...
+
+class SearchBreadcrumb implements BreadcrumbBuilderInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applies(RouteMatchInterface $route_match) {
+    $match = FALSE;
+
+    if ($route_match->getRouteName() == 'view.search.page_1') {
```
omahm: I'd be tempted to pass the required matches in as params rather than restrict it hardcoded values in the code.

johangant: Just want to check:

> pass the required matches in as params

Were you thinking of config, or something else? Quite likely I'm brain-farting and missing something in plain-sight :)

omahm: Yeah, pass them in as params in your services.yml file.

johangant: Ah, cool. I haven't ever needed to use those before, that's pretty handy. Nice one.

I've pushed up a fresh commit with the change

PR #17, opened 2019-08-22
D8 nid 280

DuttonMa:
A start on testing the workflow process, access to views being tested here.
More to come.

### DIFF
--- a/nidirect_workflow/tests/src/Functional/WorkflowAuditTest.php
+++ b/nidirect_workflow/tests/src/Functional/WorkflowAuditTest.php
@@ -9,7 +9,7 @@ use Drupal\Tests\BrowserTestBase;
  *
  * @group nidirect_common
  */
-class WorkflowTest extends BrowserTestBase {
+class WorkflowAuditTest extends BrowserTestBase {
 
   /**
    * Use install profile so that we have all content types, modules etc.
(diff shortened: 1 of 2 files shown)


johangant:
Looks like it'll do what it says. Is there a way we can run this as a Kernel test rather than a browsertest? I'm not sure we need the UI at all; perhaps setting any workflow states programmatically to test the states/outcomes of each step?

Also SimpleTest browsertests cause a bunch of issues with permissions on CircleCI which I've tried to dodge by delegating any UI tests to nightwatch.js.

Thoughts?

omahm:
> Thoughts?
True, if it's gonna be a frontend test, go for nightwatchjs. Could try as kernel tests with states but it that testing more of the end outcome, do we also want to test that they are receiving the 403?

DuttonMa:
Yes, I did think about this and was reluctant to go down the Browser test route myself, but this was my thinking:

- Kernel tests would be great but the basic thing that I am trying to test here is whether a user with a certain role/permission can see a certain page/view. I couldn't think of a way to do this in a Kernel test but any suggestions as to how I can achieve something similar would be very welcome

- Nightwatch tests - fundamentally I can't help thinking that this would be the best way to go, as really what we are talking about here is "if I log on as this user, what can I see/do ?" The main reason that I haven't (yet) gone down the Nightwatch route was that we would need to pre-create a set of users with certain roles/permissions and then log on as those users. (Unless we added those users to the live site.) Can either of you think of a neat way to achieve this?

So yes, bottom line is that I am not delighted with what I have produced! A side issue seems to be that I have created a view ('needs review') which should only be visible to certain roles. No matter how I tried in a Browser test I did not seem to be able to create a user, give them a role, and then log in as that user and access the view (it was always a 403, and really shouldn't have been).

omahm:
Closing as this will be moved to the 'origins modules'

PR #18, opened 2019-08-24
D8 nid 290 social sharing tools

johangant:
Simple module to expose a block that can be put into the menu/footer as needed, rather than having to use Drupal's menu system to do precisely the same thing with less control

### DIFF
--- a/nidirect_sst/nidirect_sst.module
+++ b/nidirect_sst/nidirect_sst.module
@@ -1,6 +1,0 @@
-<?php
-
-/**
- * @file
- * Contains nidirect_sst.module.
- */
--- a/nidirect_sst/nidirect_sst.module
+++ b/nidirect_sst/nidirect_sst.module
@@ -1,0 +1,6 @@
+<?php
+
+/**
+ * @file
+ * Contains nidirect_sst.module.
+ */
--- a/nidirect_sst/src/Plugin/Block/SharingLinks.php
+++ b/nidirect_sst/src/Plugin/Block/SharingLinks.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Drupal\nidirect_sst\Plugin\Block;
+
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Link;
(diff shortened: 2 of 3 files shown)


nidirect_sst/nidirect_sst.module
```diff
@@ -0,0 +1,6 @@
+<?php
+
+/**
+ * @file
+ * Contains nidirect_sst.module.
+ */
```
DuttonMa: As this file is empty, maybe you should just get rid of it ?

nidirect_sst/src/Plugin/Block/SharingLinks.php
```diff
@@ -0,0 +1,47 @@
...
+   * {@inheritdoc}
+   */
+  public function build() {
+    $build = [];
+
+    // Links for: Twitter, Facebook, Youtube and RSS.
+    $build['twitter'] = Link::fromTextAndUrl($this->t('Twitter'),
+      Url::fromUri('https://twitter.com/nidirect', [
+        'attributes' => ['title' => $this->t('Twitter')]
+      ]))->toRenderable();
```
DuttonMa: I like this construction - I must start building my links this way !

johangant:
Parking this PR for the moment. @omahm suggested some Twig template links as a first step, with the possibility of config from the theme layer if it needs to be re-used or re-configured with minimal fuss.

nidirect_sst/nidirect_sst.module
johangant: Agreed, didn't spot this. Thanks

johangant:
So it's entirely possible to add the render array elements from the block implementation directly into the theme region render array, although Bartik's default implementation is a tad simplistic (prints a pre-rendered HTML variable) but the nidirect theme could easily render `content.sst_links` or use preprocess_page to squirt in the relevant render elements.

I'm double-checking consensus on how best to present this stuff to avoid having to unpick/redo later.

johangant:
Marking as prospective future enhancement and closing.

PR #19, opened 2019-08-24
D8NID-282 Add custom module to control display of sidebar block view.

johangant:
This PR provides a module that gives a block to show the current position in the site themes taxonomy vocab, based on values from the node being viewed.

### DIFF
--- a/nidirect_related_content/nidirect_related_content.module
+++ b/nidirect_related_content/nidirect_related_content.module
@@ -5,19 +5,13 @@
  * Contains nidirect_related_content.module.
  */
 
-use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\views\ViewExecutable;
-use Drupal\Core\Block\BlockPluginInterface;
 
 /**
  * Implements hook_views_pre_build().
  */
 function nidirect_related_content_views_pre_build(ViewExecutable $view) {
-  if ($view->id() != 'articles_by_term') {
-    return;
-  }
-
-  if ($view->current_display != 'block_1') {
+  if ($view->id() != 'articles_by_term' && $view->current_display != 'block_1') {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


nidirect_related_content/nidirect_related_content.module
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Implements hook_views_pre_build().
+ */
+function nidirect_related_content_views_pre_build(ViewExecutable $view) {
+  if ($view->id() != 'articles_by_term') {
+    return;
+  }
+
+  if ($view->current_display != 'block_1') {
```
omahm: I'd be tempted to put this in as an && condition to the if statement above as a blanket condition for the correct View and Display.

nidirect_related_content/nidirect_related_content.module
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Implements hook_views_pre_render().
+ */
+function nidirect_related_content_views_pre_render(ViewExecutable $view) {
+  if ($view->id() != 'articles_by_term') {
+    return;
+  }
+
+  if ($view->current_display != 'block_1') {
```
omahm: As per previous comment.

nidirect_related_content/nidirect_related_content.module
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Implements hook_preprocess_views_view_field().
+ */
+function nidirect_related_content_preprocess_views_view_field(&$variables) {
+  if ($variables['view']->id() != 'articles_by_term') {
+    return;
+  }
+
+  if ($variables['view']->current_display != 'block_1') {
```
omahm: as per previous comment.

nidirect_related_content/nidirect_related_content.module
johangant: Probably more concise, agreed. I'll crunch them down a bit as suggested 👍

PR #22, opened 2019-09-03
D8NID-184 related recipes blocks

johangant:
PR to cover the definition of various sidebar blocks that generate facet URLs based on recipe node field values.

### DIFF
--- a/nidirect_related_content/src/Plugin/Block/RecipesByCourseBlock.php
+++ b/nidirect_related_content/src/Plugin/Block/RecipesByCourseBlock.php
@@ -119,6 +119,12 @@ class RecipesByCourseBlock extends BlockBase implements ContainerFactoryPluginIn
       foreach ((array) $course as $value) {
         $term_id = $value['target_id'];
         $term = $this->entityTypeManager->getStorage('taxonomy_term')->load($term_id);
+
+        // Skip if null.
+        if (empty($term)) {
+          continue;
+        }
+
         $facet_url = $this->facetsUrlGenerator->getUrl(['course_type' => [$term_id]]);
         $build['facet_link_course']['#items'][] = Link::fromTextAndUrl($term->label(), $facet_url)->toRenderable();
       }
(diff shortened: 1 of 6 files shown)


nidirect_related_content/src/Plugin/Block/RecipesByCourseBlock.php
```diff
@@ -0,0 +1,134 @@
...
+
+      // For each entry, create a new Link object and add to the render array.
+      $build['facet_link_course'] = [
+        '#theme' => 'item_list',
+        '#list_type' => 'ul',
+      ];
+      foreach ((array) $course as $value) {
+        $term_id = $value['target_id'];
+        $term = $this->entityTypeManager->getStorage('taxonomy_term')->load($term_id);
+        $facet_url = $this->facetsUrlGenerator->getUrl(['course_type' => [$term_id]]);
```
DuttonMa: Should there be some checks here? Might $term be NULL ? (In which case $term->label() will fail)
Might $term_id be NULL?

nidirect_related_content/src/Plugin/Block/RecipesByCourseBlock.php
```diff
@@ -0,0 +1,134 @@
...
+        $build['facet_link_course']['#items'][] = Link::fromTextAndUrl($term->label(), $facet_url)->toRenderable();
+      }
+
+      // Set cache context to vary by node path to avoid duplication across different pages.
+      $build['facet_link_course']['#cache']['contexts'] = ['url.path'];
```
DuttonMa: Like this caching work !

nidirect_related_content/src/Plugin/Block/RecipesByCourseBlock.php
johangant: Not a bad idea to be a little more cautious when handling this, good shout.

johangant:
> I'd have been tempted to roll them into one with a config option to select the 'by' option

I did consider this but felt it was likely a step too far for what was needed at the moment. Not a problem to revisit/abstract a little further in future if it's needed for any reason.

PR #25, opened 2019-09-06
D8NID-192 news

johangant:
Opted to use two views embed displays managed with a controller for maximum flexibility (ie: if you want to introduce something above/below/between the content lists) that might traditionally be harder on just a views page display. This also mirrors the same approach taken for contacts which had some slightly more bespoke requirements.


johangant:
Sorry for spamming with reviewers: would be useful to get @neilblair and @David-Corbett to see how this is being done.

NB: this isn't the only way this feature can be built and perhaps isn't the most conventional approach, but it's really simple to understand (I hope!), flexible to adapt and doesn't tie you in very hard to views itself - just uses the best parts for surfacing and formatting the content to save on faffing with entity type manager queries and deeper render array work.

PR #26, opened 2019-09-06
D8NID-301 ~ Fixing undefined index bug on layout builder edit screen.

omahm:


omahm:
> Confirmed: this works fine. Plugin ID a much more robust thing to check, nice one

Cheers, thanks for checking.

PR #28, opened 2019-09-16
D8NID-194 GP practice preprocessing on search results and breadcrumb

johangant:
@DuttonMa a quick Q for you: the way I'm rendering fields through to my Twig template makes me think it's not quite right? Something you mentioned in your caching discussion made me think that being selective on specific field values misses the caching metadata, did I remember that right?

### DIFF
--- a/nidirect_gp/nidirect_gp.module
+++ b/nidirect_gp/nidirect_gp.module
@@ -5,6 +5,9 @@
  * Contains nidirect_gp.module.
  */
 
+use Drupal\Component\Utility\Xss;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
 
 /**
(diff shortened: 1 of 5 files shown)


omahm:
> @DuttonMa a quick Q for you: the way I'm rendering fields through to my Twig template makes me think it's not quite right? Something you mentioned in your caching discussion made me think that being selective on specific field values misses the caching metadata, did I remember that right?

Problem is you're not gonna have the cache tags. You need to render the content array to include these. One trick is to call it but using the without filter.

nidirect_gp/nidirect_gp.module
```diff
@@ -33,3 +33,62 @@ function nidirect_gp_theme() {
...
+    return;
+  }
+
+  // Rejig the content render array based on what was presented from underlying GP entities
+  // and existing GP Practice field values.
+  $gp_details = [];
+
+  // Lead GP.
+  if (!empty($node->field_gp_practice_lead->referencedEntities())) {
+    $lead_gp = $node->field_gp_practice_lead->referencedEntities()[0];
```
omahm: You'd be better off rendering a display mode instead of creating the 'display' in code.
Same goes for the next few if statements

nidirect_gp/nidirect_gp.module
```diff
@@ -33,3 +33,62 @@ function nidirect_gp_theme() {
...
+  $variables['content'] = $gp_details;
+}
+
+/**
+ * Implements hook_preprocess_block().
+ */
+function nidirect_gp_preprocess_block(&$variables) {
+  if ($variables['plugin_id'] == 'views_exposed_filter_block:gp_practices-find_a_gp') {
+    $variables['content']['preamble'] = [
+      '#markup' => '<p>' . t('Enter a GP name, practice or town.') . '</p>',
```
omahm: I can see DC and Neil possibly looking this in a twig template.

nidirect_gp/templates/node--gp-practice--teaser.html.twig
```diff
@@ -0,0 +1,35 @@
...
+#}
+{%
+  set classes = [
+  view_mode ? 'node--view-mode-' ~ view_mode|clean_class,
+  'clearfix',
+]
+%}
+<h3><a href="{{ url }}">{{ label }}</a></h3>
+<div class="gp-practice-members">
+  {% if content.lead_gp %}
```
omahm: Need to render the content array to ensure the cache tags bubble up.

nidirect_gp/nidirect_gp.module
johangant: Noted, yep. I need to think about this more, but essentially I have a display mode for teaser which this node type uses but the field formatter options don't quite give me what I need.

I have more time than expected (!!) for this so what I can do next is:

- View mode for GP entity to present the first and last name to reduce the swapping out of existing items in the existing content section.
- I still need to render address components selectively in the Twig template, but that's easy to do when combined with the without filter.

omahm: Just create a pseudo field called 'full name' and use that in the display mode.

johangant: 👍

nidirect_gp/templates/node--gp-practice--teaser.html.twig
johangant: Thanks. Now sorted with final call to render 'content' passed via the `without` filter where some fields have been selectively used.

DuttonMa:
> @DuttonMa a quick Q for you: the way I'm rendering fields through to my Twig template makes me think it's not quite right? Something you mentioned in your caching discussion made me think that being selective on specific field values misses the caching metadata, did I remember that right?

It's OK Johan, your code is fine because it includes this line:

{{ content | without('field_address', 'field_gp_appointments', 'field_gp_prescriptions') }}

... and that will cause the metadata to be bubbled up correctly.

The problem only occurs when we just render specific fields from 'content' without rendering 'content' itself. So if you had just rendered 'content.field_address' and nothing else, that could have caused trouble.

As long as you render 'content' in full, or 'content | without' then all will be fine.

nidirect_gp/nidirect_gp.module
johangant: Good idea. I've tweaked this so it works as a template; they can copy the default version into the theme whenever they're ready to use it.

johangant:
All flagged issues (blocking and non-critical) now addressed. I'll force the merge and save on a handful of clicks for @omahm :)

PR #29, opened 2019-09-20
Add block wrapper around views embed display

johangant:
Necessary to allow exposed filter to work correctly. Normally you can add a views block display and the AJAX filters work for your needs, but because we introduce search_api_autocomplete we want to either move the user on form submit to the URL of the item selected or, if nothing, to the default form action path. AJAX filters don't allow this, so with a small block to wrap the display we can get around this limitation.


johangant:
> Can you add into the class your notes as to why this was required.

@omahm good shout, this is now in place. You may need to re-review via GH for me to merge this the standard way.

PR #37, opened 2019-10-02
D8NID-245 Fix breadcrumb path for health conditions search

johangant:
Also pegged recipe search breadcrumb to use view route name rather than
fixed URL from expected string value.

### DIFF
--- a/nidirect_breadcrumbs/src/HealthConditionBreadcrumb.php
+++ b/nidirect_breadcrumbs/src/HealthConditionBreadcrumb.php
@@ -99,7 +99,7 @@ class HealthConditionBreadcrumb implements BreadcrumbBuilderInterface {
     // Add cache tags so that if any entities above change, we can regenerate the breadcrumb too.
     $breadcrumb->addCacheTags([
       'taxonomy_term:22',
-      'node:7387'
+      'node:7387',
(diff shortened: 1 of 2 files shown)


nidirect_breadcrumbs/src/HealthConditionBreadcrumb.php
```diff
@@ -76,15 +88,18 @@ public function build(RouteMatchInterface $route_match) {
...
+    }
 
     $breadcrumb->setLinks($links);
     $breadcrumb->addCacheContexts(['url.path']);
 
     // Add cache tags so that if any entities above change, we can regenerate the breadcrumb too.
     $breadcrumb->addCacheTags([
       'taxonomy_term:22',
-      'node:7387',
+      'node:7387'
```
omahm: Does phpcs not complain about the missing comma?

johangant: Apparently not! https://www.drupal.org/docs/develop/standards/coding-standards#array

That's crept in from me fiddling with cache tags, so I'll reinstate the trailing comma for consistency.

DuttonMa:
LGTM

PR #42, opened 2019-10-22
Site themes tree

DuttonMa:
Create a 'site themes' display to replace the 'site topics' tree that was available in the D7 version of NIDirect

### DIFF
--- a/nidirect_site_themes/src/Controller/SiteThemeController.php
+++ b/nidirect_site_themes/src/Controller/SiteThemeController.php
@@ -5,7 +5,6 @@ namespace Drupal\nidirect_site_themes\Controller;
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Link;
 use Drupal\Core\Url;
-use Drupal\taxonomy\Entity\Term;
 
 /**
  * Class SiteThemeController.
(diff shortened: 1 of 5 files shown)


nidirect_site_themes/src/Controller/SiteThemeController.php
```diff
@@ -0,0 +1,74 @@
...
+    $output .= "</ul></div>";
+    return $output;
+  }
+
+  /**
+   * Display one level of tree.
+   */
+  protected function printOneLevel($vid, $parent_tid) {
+    $output = "<div class='theme_item_list'><ul>";
+    $term = Term::load($parent_tid);
```
omahm: This class has access to an injected entityTypeManager, the term should be loaded via that.

nidirect_site_themes/src/Controller/SiteThemeController.php
```diff
@@ -0,0 +1,74 @@
...
+      '#type' => 'markup',
+      '#markup' => $output,
+    ];
+  }
+
+  /**
+   * Display entire tree for one vocabulary.
+   */
+  protected function printVocab($vid) {
+    $output = "<div class='item_list'><ul>";
```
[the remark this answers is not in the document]
DuttonMa: Good idea, have done that now and have got rid of a load of code :-)

nidirect_site_themes/src/Controller/SiteThemeController.php
DuttonMa: That's fixed now too

PR #43, opened 2019-10-25
D8 nid 347 publications search

johangant:


nidirect_breadcrumbs/src/PublicationBreadcrumb.php
```diff
@@ -0,0 +1,83 @@
...
+  public function applies(RouteMatchInterface $route_match) {
+    $match = FALSE;
+    $this->node = $route_match->getParameter('node');
+
+    if (!empty($this->node)) {
+      if (is_object($this->node) == FALSE) {
+        $this->node = $this->entityTypeManager->getStorage('node')->load($this->node);
+      }
+
+      $match = $this->node->bundle() == 'publication';
```
omahm: What if we don't get a node returned from ETM and try calling bundle() on it, presumably it'll throw a method not found error.

johangant: Good shout. I see the other handlers do the same, so I'll move this line into the block above and make this more defensive.

johangant: Introduced a slightly clearer approach to this. I expect they'll evolve again as the project expands. I'd be inclined to move things like search page route names in as service parameters and maybe restructure the scope of the handlers a little but happy to tackle that once things become an issue rather than ahead of the time.

PR #44, opened 2019-11-12
DNID-350 content backlinks

johangant:
Early PR. A few minor things left to sort out, but broadly:

- Need to refactor to remove `nidirect_` prefixes to make it more friendly for release as contrib module.


nidirect_backlinks/src/Command/ScanCommand.php
```diff
@@ -0,0 +1,59 @@
...
+    if (!$db_conn->schema()->tableExists('nidirect_backlinks')) {
+      return 3;
+    }
+
+    $query = $db_conn->query('select nid from {node}');
+    $results = $query->fetchAll();
+
+    // Potentially long running task here; may need batching.
+    foreach ($results as $row) {
+      $node = \Drupal::service('entity_type.manager')->getStorage('node')->load($row->nid);
```
omahm: You're gonna be calling getStorage() on every iteration, I'd look at assigning this outside the loop and possibly injecting it as a dependency.

nidirect_backlinks/src/Command/ScanCommand.php
```diff
@@ -0,0 +1,59 @@
...
+
+      if ($node instanceof NodeInterface) {
+        \Drupal::service('nidirect_backlinks.linkmanager')->processEntity($node);
+        // $this->getIo()->info('Processed node ID: ' . $row->nid);
```
omahm: Remove commented out code.

nidirect_backlinks/nidirect_backlinks.module
```diff
@@ -0,0 +1,51 @@
...
+      return $output;
+
+    default:
+  }
+}
+
+/**
+ * Implements hook_node_insert().
+ */
+function nidirect_backlinks_node_insert(EntityInterface $entity) {
```
omahm: Is there an event you can subscribe to for these, or can it only be done with hooks?

johangant:
Good comments, and now reflected in the repo at https://git.drupalcode.org/project/whatlinkshere

I'll close here as this PR will become a composer and config change to get this working.

PR #46, opened 2019-11-18
D8NID-355 Add related conditions block plugin

johangant:
- Strip field-based preprocess code on search results; replace with rendered block plugin contents for consistency

### DIFF
--- a/nidirect_health_conditions/src/Plugin/Block/RelatedConditionsBlock.php
+++ b/nidirect_health_conditions/src/Plugin/Block/RelatedConditionsBlock.php
@@ -98,7 +98,7 @@ class RelatedConditionsBlock extends BlockBase implements ContainerFactoryPlugin
     }
 
     // But check it's what we're expecting anyway.
-    if (!empty($this->node)) {
+    if (!empty($this->node) && $this->node->hasField('field_parent_condition') && $this->node->hasField('field_related_conditions')) {
       // Load the parent condition field first.
       $parent_condition_nid = $this->node->get('field_parent_condition')->target_id;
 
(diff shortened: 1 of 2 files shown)


nidirect_health_conditions/src/Plugin/Block/RelatedConditionsBlock.php
```diff
@@ -0,0 +1,146 @@
...
+    // in contexts where this is available.
+    if ($this->node instanceof NodeInterface == FALSE) {
+      $this->node = $this->routeMatch->getParameter('node');
+    }
+
+    // But check it's what we're expecting anyway.
+    if (!empty($this->node)) {
+      // Load the parent condition field first.
+      $parent_condition_nid = $this->node->get('field_parent_condition')->target_id;
+
```
DuttonMa: Will this fail elegantly if the field doesn't exist on this node?

nidirect_health_conditions/src/Plugin/Block/RelatedConditionsBlock.php
```diff
@@ -0,0 +1,146 @@
...
+        $parent_condition = $storage->load($parent_condition_nid);
+        $links[] = Link::createFromRoute(
+          $parent_condition->label(),
+          'entity.node.canonical',
+          ['node' => $parent_condition_nid]
+        );
+      }
+
+      // Get related conditions and add to links array if we have values.
+      $related_conditions = $this->node->get('field_related_conditions')->getValue();
```
DuttonMa: Will this fail elegantly if the field doesn't exist on this node?

johangant:
> Nice one 👍
> Only thing I was tempted to comment on was for a has field check against the parent and related conditions fields.

No harm in that in case the data model ever changes in future. I'll pop that in and then merge away 👍

PR #51, opened 2019-11-29
D8 nid 367

DuttonMa:
Add hooks to invalidate custom cache tags for taxonomy term view pages.

See other associated PRs:

- https://github.com/dof-dss/nicsdru_nidirect_theme/pull/19/files (preprocess function)
- https://github.com/dof-dss/nidirect-drupal/pull/107 (new views)

### DIFF
--- a/nidirect_common/nidirect_common.module
+++ b/nidirect_common/nidirect_common.module
@@ -222,6 +222,22 @@ function nidirect_common_node_delete(EntityInterface $entity) {
 }
 
 /**
+ * Implements hook_ENTITY_TYPE_presave() for taxonomy_term entities.
+ */
+function nidirect_common_taxonomy_term_presave(EntityInterface $entity) {
+  // Invalidate 'taxonomy_term_list' custom cache tag for the
+  // parent when a new taxonomy term is created.
+  if ($entity->get('vid')->target_id == 'site_themes') {
+    if ($entity->isNewRevision()) {
+      $parent = $entity->get('parent')->target_id;
+      if ($parent) {
+        Cache::invalidateTags(['taxonomy_term_list:' . $parent]);
+      }
+    }
+  }
+}
+
+/**
  * Invalidate taxonomy cache tags.
  */
 function _invalidate_theme_cache_tags($entity) {
--- a/nidirect_common/nidirect_common.module
+++ b/nidirect_common/nidirect_common.module
@@ -207,7 +207,6 @@ function nidirect_common_node_view(array &$build, EntityInterface $entity, Entit
 function nidirect_common_node_presave(EntityInterface $entity) {
   // Check that the node is published.
   if ($entity->get('status')->value) {
-    //_invalidate_theme_cache_tags($entity);
     $cache_service = \Drupal::service('nidirect_common.invalidate_taxonomy_list_cache_tags');
     $cache_service->invalidateForEntity($entity);
   }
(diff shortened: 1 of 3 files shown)


nidirect_common/nidirect_common.module
```diff
@@ -198,3 +198,70 @@ function nidirect_common_node_view(array &$build, EntityInterface $entity, Entit
...
+
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * Invalidate taxonomy cache tags after node save.
+ */
+function nidirect_common_node_presave(EntityInterface $entity) {
+  // Check that the node is published.
+  if ($entity->get('status')->value) {
+    _invalidate_theme_cache_tags($entity);
```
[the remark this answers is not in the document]
DuttonMa: Good idea, have done that now

nidirect_common/src/InvalidateTaxonomyListCacheTags.php
```diff
@@ -0,0 +1,42 @@
...
+        if (isset($entity->original)) {
+          $tid = $entity->original->get($thisfield)->target_id;
+          if (isset($tid)) {
+            $taxonomy_tags[] = 'taxonomy_term_list:' . $tid;
+          }
+        }
+      }
+    }
+    if (count($taxonomy_tags) > 0) {
+      Cache::invalidateTags($taxonomy_tags);
```
johangant: How about injecting the `cache_tags.invalidator` service from the container?

nidirect_common/src/InvalidateTaxonomyListCacheTags.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Class InvalidateTaxonomyListCacheTags.
+ */
+class InvalidateTaxonomyListCacheTags {
+
+  /**
+   * Invalidate custom cache tags for this entity.
+   */
+  public function invalidateForEntity($entity) {
```
johangant: A type hint of `EntityInterface` helps here

nidirect_common/src/InvalidateTaxonomyListCacheTags.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Drupal\nidirect_common;
+
+use Drupal\Core\Cache\Cache;
+
+/**
+ * Class InvalidateTaxonomyListCacheTags.
+ */
+class InvalidateTaxonomyListCacheTags {
```
johangant: Now this class has moved to a service, you should be able to inject the cache_tags.invalidator service as an argument in the services.yml file and wire it up to an internal class variable for use in any functions through the usual constructor + static create functions.

nidirect_common/src/InvalidateTaxonomyListCacheTags.php
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Constructs a new \Drupal\Core\Menu\MenuTreeStorage.
+   *
+   * @param \Drupal\Core\Cache\CacheTagsInvalidatorInterface $cache_tags_invalidator
+   *   The cache tags invalidator.
+   */
+  public function __construct(CacheTagsInvalidatorInterface $cache_tags_invalidator) {
+    $this->cacheTagsInvalidator = $cache_tags_invalidator;
```
johangant: Argh, PHP! Does this not grumble about the class variable not existing before assignment in the constructor?

TL;DR: would expect to see a `protected $cacheTagsInvalidator;` class variable in there. If this works okay at runtime my guess would be that PHP assigns it on-the-fly which is both handy but also a bit frustrating for consistency :)

DuttonMa: Ah good spot, will fix it in the next PR as I am working in there again anyway

PR #55, opened 2019-12-04
custom description for theme/subtheme field on landing page nodes

DuttonMa:


nidirect_common/nidirect_common.module
```diff
@@ -237,3 +237,18 @@ function nidirect_common_taxonomy_term_presave(EntityInterface $entity) {
...
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function nidirect_common_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Add description to theme/subtheme field for landing pages.
+  if (($form_id == 'node_landing_page_form') || ($form_id == 'node_landing_page_edit_form')) {
+    if (isset($form['field_subtheme'])) {
+      $msg = "Note that the theme/subtheme that you choose here will be overridden by this landing page. <br/>";
```
johangant: Great to see this here. An non-blocking PR enhancement would be to wrap this message in `t()` so it becomes translatable.

nidirect_common/nidirect_common.module
```diff
@@ -237,3 +237,18 @@ function nidirect_common_taxonomy_term_presave(EntityInterface $entity) {
     }
   }
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function nidirect_common_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Add description to theme/subtheme field for landing pages.
+  if (($form_id == 'node_landing_page_form') || ($form_id == 'node_landing_page_edit_form')) {
```
johangant: I was going to suggest using route names instead of form IDs (eg: entity.node.edit_form) as they feel more natural in D8, but this is a form alter hook and the value is already provided, so I think a route based check would be less efficient.

nidirect_common/nidirect_common.module
DuttonMa: It is already done in line 251

PR #58, opened 2019-12-11
D8-NID 242 driving instructor breadcrumbs

johangant:
Also includes some smaller fixes for existing breadcrumb handlers + nightwatch tests for them.


nidirect_breadcrumbs/src/DrivingInstructorsBreadcrumb.php
```diff
@@ -0,0 +1,102 @@
...
+   */
+  public function build(RouteMatchInterface $route_match) {
+    $breadcrumb = new Breadcrumb();
+    $term_storage = $this->entityTypeManager->getStorage('taxonomy_term');
+
+    $links = [];
+    // Home.
+    $links[] = Link::createFromRoute(t('Home'), '<front>');
+    // Motoring term (12).
+    $term = $term_storage->load(12);
```
omahm: Instead of individually loading, why not use loadMultiple() ?

johangant: Opted for a very literal approach in the first instance. Latest changes now have a loadMultiple + loop to build up the links.

PR #60, opened 2019-12-20
Article teaser views

DuttonMa:
Created a custom block to display article teasers on campaign / landing pages.
Two view displays that return teasers are rendered and then sorted before being displayed.

See related PRs:
https://github.com/dof-dss/nidirect-drupal/pull/132
https://github.com/dof-dss/nicsdru_nidirect_theme/pull/29


nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
```diff
@@ -0,0 +1,157 @@
...
+    $build['nidirect_article_teasers_by_topic']['#cache'] = [
+      'tags' => $cache_tags,
+    ];
+    return $build;
+  }
+
+  /**
+   * Utility function to render 'articles by term' view.
+   */
+  private function renderArticleTeasersByTerm($tid, $current_nid, &$cache_tags) {
```
johangant: Not critical, but if you put type hints before each parameter it helps autodoc tools and PHPStorm to detect any inconsistencies before runtime. I'd suggest something like `(int $tid, int $current_nid, array &$cache_tags)`

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
```diff
@@ -0,0 +1,157 @@
...
+      // Add cache tag for each article.
+      $cache_tags[] = 'node:' . $row->nid;
+    }
+    return $results;
+  }
+
+  /**
+   * Utility function to render 'site subtopics' view.
+   */
+  private function renderArticleTeasersByTopic($tid, &$cache_tags) {
```
johangant: Type hints like `(int $tid, int &$cache_tags)` would be good

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
DuttonMa: Arg! That's one that you always spot - and one that I always forget !
Good point, done now

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
DuttonMa: Done now

PR #63, opened 2019-12-23
D8 nid 169

omahm:


nidirect_webforms/templates/nidirect-webforms-quiz-answer-feedback.html.twig
```diff
@@ -0,0 +1,9 @@
+{% set classes = [
+  passed ? 'success-notice' : 'error-notice'
+]
+%}
+<div {{ attributes.addClass(classes) }}>
+  <h2>{{ title }}</h2>
+  <p class="question">{{ question }}</p>
+  <div class="feedback"> {{ feedback }}</div>
```
DuttonMa: don't think there should be a space before {{ feedback }} ?

omahm: Good spot!

PR #67, opened 2020-01-07
D8NID 188 news listing

johangant:


nidirect_news/src/Controller/NewsListingController.php
```diff
@@ -71,6 +74,11 @@ public function default() {
       '#display_id' => 'latest_news',
     ];
 
+    // View title: see views_embed_view() which the render array relies on for details of why this is missing.
+    $content['older_news_title'] = [
+      '#markup' => '<h2 class="hr-above">' . t('Older news items') . '</h2>',
```
omahm: I'd look at returning this as a render array of type html_tag

johangant: Perfect. This was something of a stopgap as I couldn't recall the precise thing needed towards the end of the day, that's the one I need though... cheers.

PR #73, opened 2020-01-12
D8 nid 398

omahm:


nidirect_driving_instructors/nidirect_driving_instructors.module
```diff
@@ -37,6 +37,29 @@ function nidirect_driving_instructors_form_alter(&$form, &$form_state, &$form_id
   if ($form['#id'] == 'views-exposed-form-driving-instructors-driving-instructor-results') {
     $form['sort_by']['#access'] = FALSE;
     $form['sort_order']['#access'] = FALSE;
+
+    // Add additional form classes.
+    $form['#attributes']['class'][] = 'search-form';
+    $form['categories']['#attributes']['class'][] = 'container-inline';
+
+    // Add postcode placeholder example.
+    $form['search_api_fulltext']['#placeholder'] = "e.g. 'BT1'";
```
johangant: Wrap in `t()`?

omahm: Left it out because I didn't really think it was something to be translated, what are your thoughts @DuttonMa ?

DuttonMa: Well it seems pretty unlikely, but I suppose that we should go ahead and make things translatable wherever possible

PR #74, opened 2020-01-13
D8 nid 407

omahm:


nidirect_health_conditions/src/Plugin/Block/HealthConditionsAzBlock.php
```diff
@@ -85,6 +85,12 @@ public function build() {
       }
     }
 
+    $build['healthconditions_az_block_intro'] = [
+      '#type' => 'html_tag',
+      '#tag' => 'p',
+      '#value' => 'Or find conditions beginning with ...'
```
johangant: Need an encapsulating `t()` for this?

omahm: DoH!

PR #80, opened 2020-01-16
Contact views

DuttonMa:
Create temporary module to provide a route/controller for the Government Departments Contact page (will later be replaced by a 'Page' node)

This module is used by https://github.com/dof-dss/nidirect-drupal/pull/170

### DIFF
--- a/nidirect_common/nidirect_common.links.menu.yml
+++ b/nidirect_common/nidirect_common.links.menu.yml
@@ -20,7 +20,11 @@ nidirect.main.help:
   weight: 7
 nidirect.main.feedback:
   title: 'Feedback'
-  url: internal:/form/site-feedback
+  route_name: entity.webform.canonical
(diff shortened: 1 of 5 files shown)


nidirect_common/nidirect_common.links.menu.yml
```diff
@@ -20,7 +20,7 @@ nidirect.main.help:
   weight: 7
 nidirect.main.feedback:
   title: 'Feedback'
-  url: internal:/forms/feedback?s=/front
+  url: internal:/form/site-feedback
```
johangant: Worth checking to see if they still want the `?s=/front` parameter in case it's relevant for GA tracking of some sort

DuttonMa: Good thought, I checked with Neil and he did want the parameter preserved.
Have now used the canonical path in there, please re-review

PR #85, opened 2020-01-30
Fixes a-z facet label to be an H2 and added aria-describedby to the f…

neilblair:
…acet UL to point to the H2's ID. Also removed title attributes from the facet links.


johangant:
Had the branch in front of me so merged/pushed accordingly after the PR was approved 👍

PR #89, opened 2020-01-30
D8NID-433 Check for gp_names preprocess variable before trying to manipulate it

johangant:


nidirect_gp/nidirect_gp.module
```diff
@@ -77,7 +77,7 @@ function nidirect_gp_preprocess_node(&$variables) {
   if ($variables['node']->bundle() == 'gp_practice' && $variables['view_mode'] == 'search_result') {
     // Prune a trailing '#weight' item key in the GP names array which is added
     // By something after hook_entity_type_view() above.
-    if (array_key_exists('#weight', $variables['content']['gp_names'])) {
+    if (!empty($variables['content']['gp_names']) && array_key_exists('#weight', $variables['content']['gp_names'])) {
```
DuttonMa: Should you go even further here? Are you sure that $variables['content']['gp_names'] will exist ?

johangant: Yes, this will be okay. `empty()` can handle array keys that don't exist, and the `&&` logic operator will skip the second condition if the first fails as there's no point checking the second if both are needed.

Mocked this up with a REPL example (formatting might be bit janky, sorry):

```
php > $variables['content'] = [];
php > if (!empty($variables['content']['gp_names']) && array_key_exists('#weight', $variables['content']['gp_names'])) { echo 'Success'; } else { echo 'Nope'; }
Nope
php > $variables['content']['gp_names'] = ['Yogi', 'Boo boo'];
php > if (!empty($variables['content']['gp_names']) && array_key_exists('#weight', $variables['content']['gp_names'])) { echo 'Success'; } else { echo 'Nope'; }
Nope
php > $variables['content']['gp_names']['#weight'] = 1;
php > if (!empty($variables['content']['gp_names']) && array_key_exists('#weight', $variables['content']['gp_names'])) { echo 'Success'; } else { echo 'Nope'; }
Success
```

PR #92, opened 2020-02-04
D8NID-449 ~ Wrap category facet in fieldset

omahm:


omahm:
Closing this as there are wider issues with the category selection not working and this will be addressed in D8NID-461.

PR #94, opened 2020-02-06
D8 nid 460 landing page duplicates and redirectds

johangant:
Two key things in this PR:

1. It aims to avoid creating redirects for new landing pages when they already exist.
2. It aims to warn editors when saving a new landing page whether another already exists for that theme value. That should ward off any duplicate content issues and, conversely, also reduce the likelihood of duplicate redirects being created.

There's a fair bit of refactoring to move a lot of code shared between entity insert + update into a separate .inc file for better readability.

### DIFF
--- a/nidirect_common/inc/landing_page.redirects.inc
+++ b/nidirect_common/inc/landing_page.redirects.inc
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Contains functions Layout Builder search indexing.
+ * Contains functions for handling landing page creation.
  */
 
 use Drupal\Core\Link;
@@ -77,13 +77,11 @@ function nidirect_common_is_landing_page_with_theme_value(EntityInterface &$enti
  */
 function nidirect_common_get_redirect_url(string $source_path) {
   $redirect = \Drupal::service('redirect.repository')->findBySourcePath($source_path);
+  $redirect_url = '';
 
   if (!empty(($redirect))) {
     $redirect = array_pop($redirect);
     $redirect_url = $redirect->getRedirectUrl()->toString();
-  }
-  else {
-    $redirect_url = '';
   }
 
   return $redirect_url;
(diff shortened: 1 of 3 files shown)


nidirect_common/inc/landing_page.redirects.inc
```diff
@@ -0,0 +1,129 @@
+<?php
+
+/**
+ * @file
+ * Contains functions Layout Builder search indexing.
```
omahm: Think you need to update the file description

nidirect_common/inc/landing_page.redirects.inc
```diff
@@ -0,0 +1,129 @@
...
+ */
+function nidirect_common_get_redirect_url(string $source_path) {
+  $redirect = \Drupal::service('redirect.repository')->findBySourcePath($source_path);
+
+  if (!empty(($redirect))) {
+    $redirect = array_pop($redirect);
+    $redirect_url = $redirect->getRedirectUrl()->toString();
+  }
+  else {
+    $redirect_url = '';
```
omahm: I'd assign this var at the top and then you can do away with the else statement.

nidirect_common/inc/landing_page.redirects.inc
```diff
@@ -0,0 +1,129 @@
...
+      'language' => Language::LANGCODE_NOT_SPECIFIED,
+      'status_code' => '301',
+    ])->save();
+  }
+  else {
+    // Notify the user that a redirect could not be created.
+    \Drupal::messenger()->addWarning(t('A redirect from the theme taxonomy page to this content was not created,
+        because there are already one or more redirects used for the theme taxonomy page.'));
+
+    // Add a convenience link for users who can administer redirects.
```
omahm: Like it 👍

nidirect_common/inc/landing_page.redirects.inc
johangant: *facepalm* My copy/pasta secrets must not be known :)

Good spot!

nidirect_common/inc/landing_page.redirects.inc
johangant: Yep, agreed. Like this

PR #98, opened 2020-02-07
Landing redirect bugfixes

DuttonMa:


nidirect_common/inc/landing_page.redirects.inc
```diff
@@ -28,10 +28,19 @@ use Drupal\redirect\Entity\Redirect;
...
-    ->condition('field_subtheme', $theme_selected, '=')
-    ->execute();
+  $query->condition('type', 'landing_page', '=');
+  if (!empty($current_node)) {
+    // Exclude current node.
+    $query->condition('nid', $current_node->id(), '!=');
+  }
+  $query->condition('field_subtheme', $theme_selected, '=');
+  // We are only worried about published nodes.
+  $query->condition('status', 1);
```
johangant: I originally figured you want to prevent 2 editors crossing the proverbial streams by setting up/working on two separate landing pages, but in all likelihood, this isn't going to be much of a real-world problem. I think leave this as published for now and if there's a demand for it to be less restrictive then we can relax this condition.

DuttonMa: Yes it is indeed debatable, but after doing some testing for some related work I came to the conclusion that this was the best way to do it

PR #101, opened 2020-02-07
D8 nid 427

DuttonMa:


nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Utility Ajax function.
+ */
+function nidirect_landing_pages_populate_listing(array &$form, FormStateInterface $form_state) {
+  // The main purpose of this is to make sure that
+  // nidirect_common_node_prepare_form is called.
+  if (!$form_state->getValue('field_manually_control_listing')['value']) {
+    // If 'manual control' checkbox is unchecked, make sure
+    // that listing field is not shown.
+    $form['field_listing']['#access'] = FALSE;
```
johangant: Do you want this to be 'hidden' instead? I've had a few instances where I've tried this approach and when handling the form submission I can't get at the value because '#access = FALSE' not only hides the form element but effectively removes it from the form's payload when submitted. If this works, that's great... feel free to disregard 👍

PR #104, opened 2020-02-07
D8 nid 427

DuttonMa:
Allow content editors to control the order of teasers shown on landing pages.

Currently a list of teasers is displayed at the bottom of the landing page according to the sub-theme that has been selected. These teasers are a mixture of two view displays handled by the nidirect_custom_blocks module. 

In the current implementation, these teasers always appear alphabetically. The purpose of this change is to allow the content editor to click the 'manually control listing' checkbox when editing a landing page and then to be able to take control of the teasers by re-ordering and/or removing.

This is  done by creating a paragraph field which appears when the 'manual control' checkbox is checked. The user may then manipulate the paragraphs (each of which represents a teaser) and save the end result. This then affects the teasers displayed by the nidirect_custom_blocks module.

Linked to PR https://github.com/dof-dss/nidirect-drupal/pull/199


nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
```diff
@@ -94,6 +105,30 @@ public function build() {
     return $build;
   }
 
+  /**
+   * Utility function to sort article teasers.
+   */
+  private function userSortResults($node, array $results) {
```
johangant: I'd recommend type hinting the `$node`. Something like:

```diff
- private function userSortResults($node, array $results) {
+ private function userSortResults(NodeInterface $node, array $results) {
```
With the obligatory function annotations too.

nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -42,7 +43,129 @@ function nidirect_landing_pages_entity_update(EntityInterface $entity) {
 }
 
 /**
- * Implements hook_form_alter().
+ * Implements hook_ENTITY_TYPE_prepare_form().
+ */
+function nidirect_landing_pages_node_prepare_form(EntityInterface $entity, $operation, FormStateInterface $form_state) {
```
johangant: How's this?

```diff
- function nidirect_landing_pages_node_prepare_form(EntityInterface $entity, $operation, FormStateInterface $form_state)
+ function nidirect_landing_pages_node_prepare_form(EntityInterface $entity, string $operation, FormStateInterface $form_state)
```

nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -42,7 +43,129 @@ function nidirect_landing_pages_entity_update(EntityInterface $entity) {
...
+
+/**
+ * Implements hook_page_attachments().
+ */
+function nidirect_landing_pages_page_attachments(array &$attachments) {
+  // Add landing page js if we are creating or editing
+  // a landing page node.
+  $add_js = FALSE;
+  if (\Drupal::service('router.admin_context')->isAdminRoute()) {
+    if (\Drupal::request()->server->get('REQUEST_URI') == '/node/add/landing_page') {
```
johangant: Would be more robust to use the route name, but not a blocking issue.

nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -52,9 +175,50 @@ function nidirect_landing_pages_form_alter(&$form, FormStateInterface $form_stat
...
+/**
+ * Utility Ajax function.
+ */
+function nidirect_landing_pages_populate_listing(array &$form, FormStateInterface $form_state) {
+  // The main purpose of this is to make sure that
+  // nidirect_common_node_prepare_form is called.
+  if (!$form_state->getValue('field_manually_control_listing')['value']) {
+    // If 'manual control' checkbox is unchecked, make sure
+    // that listing field is not shown.
+    $form['field_listing']['#access'] = FALSE;
```
johangant: Do you want to visually hide it or remove it from the form payload altogether? I've found this approach can cause a headache if you're using the value later on as `#access` can be too severe. If you do need it later, changing to hidden might be better.

nidirect_landing_pages/js/landing_listing.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+(function ($, Drupal) {
+  Drupal.behaviors.landinglisting = {
+    attach: (context) => {
+      // Look through all <select>s.
+      $(context).find('select').each(function() {
+        var $thisclass = $(this).attr('class');
+        // Find any whose class starts with 'shs-'.
+        if ($thisclass.indexOf('shs-') == 0) {
```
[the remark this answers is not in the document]
DuttonMa: good point, that's done now

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
DuttonMa: that's done

nidirect_landing_pages/nidirect_landing_pages.module
DuttonMa: that's done

nidirect_landing_pages/nidirect_landing_pages.module
DuttonMa: yes, good idea, have re-jigged the code to use route name now, which does feel much better

nidirect_landing_pages/nidirect_landing_pages.module
DuttonMa: very good point, have now used hidden and it does the job

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
```diff
@@ -120,6 +155,7 @@ private function renderArticleTeasersByTerm(int $tid, int $current_nid, array &$
         '#url' => Url::fromRoute('entity.node.canonical', ['node' => $row->nid]),
       ];
       $thisresult['summary_text'] = ['#markup' => $row->_entity->field_summary->value];
+      $thisresult['key'] = 'a:' . $row->nid;
```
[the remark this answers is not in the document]
DuttonMa: ok, done

nidirect_custom_blocks/src/Plugin/Block/NIDirectArticleTeasersByTopic.php
```diff
@@ -120,6 +155,7 @@ private function renderArticleTeasersByTerm(int $tid, int $current_nid, array &$
...
       // Place in an array keyed by lower case title (for sorting).
       $results[strtolower($row->_entity->getTitle())] = $thisresult;
```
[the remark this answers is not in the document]
DuttonMa: good spot, have done that now

PR #109, opened 2020-02-12
D8NID-471 ~ Hide top level theme from node forms

omahm:


omahm:
> Gah, that's annoying it won't accept a hidden field type. Good to see there's a workaround regardless

It's because it's a container with an autocomplete field, could set the nested input to hidden but then you'd end up with the label / container still visible, simplest option is just to hide the entire thing.

PR #111, opened 2020-02-14
D8NID-484 node bug fix

DuttonMa:


nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -165,9 +165,14 @@ function nidirect_landing_pages_page_attachments(array &$attachments) {
     }
     else {
       $node = \Drupal::routeMatch()->getParameter('node');
-      if (!empty($node) && ($node->getType() == 'landing_page')) {
-        // We are editing a landing page.
-        $add_js = TRUE;
+      if (!empty($node)) {
```
johangant: Could you simplify this with something like...

```
if ($node instanceof NodeInterface && $node->bundle/getType() == 'landing_page') {
... do something;
}
```

The first statement should handle nulls/other types and if that fails, the second comparison won't be executed as the expression automatically fails the logic test (needing both parts to be TRUE).

DuttonMa: Well, I like the idea but I don't think that it will work here. This is because $node can sometimes be a string which contains the nid, in which case we need to load up the node and then look at the type.
I have improved the code slightly after your comment by adding a check for NodeInterface, though.

PR #120, opened 2020-02-21
D8 nid 490 map embed ux improvements

johangant:
This PR simplifies the entity embed process for map paragraphs bundles. It's all AJAXified so a basic form_alter won't cut it: it needs to selectively pre-set a form state value so that Maps skip the embed/review steps of the entity embed browser, as well as return to the select step when editing an existing map entity.

There are some element alterations to tuck away the lat/lon coordinates into a collapsed fieldset and some title changes too.


nidirect_media/nidirect_media.module
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Implements hook_entity_embed_alter().
+ *
+ * Swap between entity embed view modes for Maps when you're viewing or editing nodes.
+ */
+function nidirect_media_entity_embed_alter(array &$build, EntityInterface $entity, array &$context) {
+  if (\Drupal::routeMatch()->getRouteName() == 'entity.node.canonical') {
+    // Dynamic iframes are prohibited in CKEditor which breaks the display of
+    // maps for editors, so we swap view mode to use a static map instead.
```
omahm: Gotta admit this is genius.

PR #122, opened 2020-03-02
Add functional JS test to check content type presence.

johangant:
- Past config import bugs (mostly from core) have meant partially
  created content types have been left 'dangling' with no means to
recreate or remove without a full reinstall.
- This check will, at the very least, allow us to detect this a little
  sooner if it does happen again.

This test will need connecting up with the Circle CI jobs/triggers to make this return any significant value. Otherwise, we're relying on devs to execute the tests which isn't a terribly reliable approach :)

### DIFF
--- a/nidirect_common/tests/src/Nightwatch/Tests/contentTypeValidation.js
+++ b/nidirect_common/tests/src/Nightwatch/Tests/contentTypeValidation.js
@@ -1,0 +1,46 @@
+const fs = require('fs');
+var types = [];
+const configDir = '../../config/sync';
+
+module.exports = {
+  '@tags': [
+    'nidirect',
+    'nidirect_common',
+  ],
+
+  'Test whether content types found in config folder are present in the UI': browser => {
+    browser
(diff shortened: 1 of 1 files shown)


nidirect_common/tests/src/Nightwatch/Tests/contentTypeValidation.js
```diff
@@ -0,0 +1,46 @@
...
+                types.push(matches[1]);
+              }
+            }
+          }
+        });
+
+        browser
+          .drupalLogin({name: process.env.TEST_USER, password: process.env.TEST_PASS})
+          .drupalRelativeURL('/admin/structure/types')
+          .waitForElementVisible('body', 1000)
```
omahm: I'd be tempted to raise this pause a little, I've had other tests fail due to body load speeds.

johangant: Yep, makes sense. It now matches the pause times in the migration modules.

PR #124, opened 2020-03-04
D8NID-532 Only show latest news + older news title when not on pager

johangant:
- Use 'page' query paramter presence/value to determine when to do this.

### DIFF
--- a/nidirect_news/src/Controller/NewsListingController.php
+++ b/nidirect_news/src/Controller/NewsListingController.php
@@ -67,22 +67,24 @@ class NewsListingController extends ControllerBase {
   public function default() {
     $content = [];
 
-    // Latest news embed display.
-    $content['latest_news'] = [
-      '#type' => 'view',
-      '#name' => 'news',
-      '#display_id' => 'latest_news',
-    ];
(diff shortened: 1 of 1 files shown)


nidirect_news/src/Controller/NewsListingController.php
```diff
@@ -67,22 +67,24 @@ public static function create(ContainerInterface $container) {
   public function default() {
     $content = [];
 
-    // Latest news embed display.
-    $content['latest_news'] = [
-      '#type' => 'view',
-      '#name' => 'news',
-      '#display_id' => 'latest_news',
-    ];
+    if (empty($this->requestStack->getCurrentRequest()->query->get('page'))) {
```
omahm: If they click on a pager link, then back to page 1 it'll have the page parameter, I think you'll need an extra check in here for page=0

johangant: Yes, I spotted that. It looks like when you have `page=0` the value is passed into `empty()` and it's the same outcome as when the parameter is missing 👍

omahm: Ahh sweet 👍

PR #128, opened 2020-03-06
D8NID-537

DuttonMa:
Add skip link to Contacts A-Z

Requires https://github.com/dof-dss/nicsdru_nidirect_theme/pull/88


nidirect_contacts/src/Plugin/Block/ContactAzBlock.php
```diff
@@ -59,16 +59,19 @@ public static function create(ContainerInterface $container, array $configuratio
...
     $build = [];
     $links = [];
+    $skip_link = '';
 
     $title = $this->t('Find contacts beginning with') . '...';
 
     if ($this->routeMatch->getRouteName() == 'nidirect_contacts.letter') {
       $letter = $this->routeMatch->getParameter('letter');
       $title = $this->t('Showing entries for :letter', [':letter' => strtoupper($letter)]);
+      $skip_link = '<a href="#contact_links" class="skip-link visually-hidden focusable">';
```
omahm: Check the use og underscore vs hyphen

PR #134, opened 2020-03-31
D8 nid 590

DuttonMa:
Override erroneous core date error message for more date fields.

### DIFF
--- a/nidirect_common/nidirect_common.module
+++ b/nidirect_common/nidirect_common.module
@@ -284,10 +284,16 @@ function nidirect_common_form_node_form_alter(&$form, FormStateInterface $form_s
     $form['field_top_level_theme']['#attributes']['class'][] = 'hidden';
   }
 
-  if (array_key_exists('field_published_date', $form)) {
-
-    // Add validation code to override core bug with date formats as
-    // described here https://www.drupal.org/project/drupal/issues/3093267
+  // Add extra date validation code to override core bug with
+  // date formats as described here
+  // https://www.drupal.org/project/drupal/issues/3093267
+  $add_date_validation = FALSE;
+  foreach ($form as $key => $field) {
+    if (preg_match('/_date$/', $key )) {
+      $add_date_validation = TRUE;
+    }
+  }
+  if ($add_date_validation) {
     array_unshift($form['#validate'], 'nidirect_common_validate_publication_date');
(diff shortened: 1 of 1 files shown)


nidirect_common/nidirect_common.module
```diff
@@ -302,20 +308,60 @@ function nidirect_common_validate_publication_date(&$form, FormStateInterface $f
...
+    'field_cwp_payments_period',
+  ];
+
   // Now loop through and re-apply all form error messages that
-  // don't involve 'field_published_date'.
+  // don't involve the specified date fields.
   foreach ($form_errors as $name => $error_message) {
-    if (substr($name, 0, 20) != 'field_published_date') {
-      $form_state->setErrorByName($name, $error_message);
+    if (preg_match('/^[a-z_]*/', $name, $matches)) {
```
johangant: `^[a-z_]*` should match any lowercase alphanumeric character at the start of the string. The `*` quantifier means zero or more of the last matching group. Does that mean this will then always be true? I'd normally consider `+` which is one or more of the last group.

nidirect_common/nidirect_common.module
```diff
@@ -302,20 +308,60 @@ function nidirect_common_validate_publication_date(&$form, FormStateInterface $f
...
+ */
+function validate_date_field(&$form_state, $date_field) {
+  // Now validate the date and raise an error
   // if it is invalid.
-  $dt = $form_state->getValue('field_published_date');
-  $date = $dt[0]['value'];
-  if (!($date instanceof DrupalDateTime) || $date->hasErrors()) {
-    // Generate error to override the erroneous core error.
-    $form_state->setErrorByName('field_published_date', t('The publication date is invalid. Please enter a date in the format dd/mm/yyyy'));
+  $dt = $form_state->getValue($date_field);
```
johangant: `$dt` == date and time?

nidirect_common/nidirect_common.module
DuttonMa: There should always be something there, so have changed this to a '+' as suggested

DuttonMa:
> Wee bit of repetition on some of the code but looks good 👍

Yes, was aware of that but decided that to remove the repeated code would make it less clear - the error message is slightly different in each case.

PR #136, opened 2020-04-02
D8 nid 617

omahm:


nidirect_common/nidirect_common.module
```diff
@@ -266,6 +267,29 @@ function nidirect_common_inline_entity_form_entity_form_alter(&$entity_form, &$f
...
+    ');
+  }
+
+  // Hide Featured Content flag.
+  if (NestedArray::keyExists($form, ['flag', 'featured_content'])) {
+    $form['flag']['hide_content']['#access'] = FALSE;
+  }
+
+  // Hide Promotion Options.
+  $form['options']['#access'] = FALSE;
```
johangant, line 284: Can we hide these via the form display settings?

omahm: Can do but it was 1 line of code Vs updating 17+ Entity Form display configurations.

PR #138, opened 2020-04-02
D8NID-220: homepage controller and featured content block

johangant:
- Provides a basic placeholder controller to replace the default view handling the site frontpage at /node.
- It also wraps a new content type (featured content list) to provide re-usable containers of feature nodes. Presently only used on the homepage but when expanded can be used elsewhere, too.
- Feature nodes, when viewed by non-privileged users will also redirect from the canonical page to the front page.


johangant:
Related to https://github.com/dof-dss/nidirect-drupal/pull/216

PR #140, opened 2020-04-03
D8NID-548

DuttonMa:
add extra form class on contact search page

Linked to https://github.com/dof-dss/nicsdru_nidirect_theme/pull/99

### DIFF
--- a/nidirect_contacts/nidirect_contacts.module
+++ b/nidirect_contacts/nidirect_contacts.module
@@ -19,6 +19,12 @@ function nidirect_contacts_form_views_exposed_form_alter(&$form, FormStateInterf
     // we still get to use URL parameters despite having no visible form elements.
     hide($form['sort_by']);
     hide($form['sort_order']);
+    // Add an extra class if this is a search results page.
+    if (!empty($form_state->getUserInput())) {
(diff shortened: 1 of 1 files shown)


nidirect_contacts/nidirect_contacts.module
```diff
@@ -19,6 +19,12 @@ function nidirect_contacts_form_views_exposed_form_alter(&$form, FormStateInterf
     // we still get to use URL parameters despite having no visible form elements.
     hide($form['sort_by']);
     hide($form['sort_order']);
+    // Add an extra class if this is a search results page.
+    if (!empty($form_state->getUserInput())) {
```
johangant, line 23: Just a stylistic point to remove one level of indentation: you could probably refine this a little with:

```
if (!empty($form_state->getUserInput()) && !empty($form_state->getUserInput()['query_contacts_az']))
```

If the first condition fails, it won't evaluate the second as it needs both expressions to be TRUE to continue.

omahm: Could even try dumping the return from that method into a var for checking and reduce the length of that statement by around 40 chars.

DuttonMa: Fair enough, have implemented those suggestions

PR #142, opened 2020-04-08
D8 nid 556

omahm:


omahm:
> Looks great. One area of possible improvement would be to encapsulate the widget field values in `t()` to allow them to be translatable, but certainly not essential at this stage.

I was going to leave the translation over to the frontend side, I don't think we need it for admin pages?

PR #151, opened 2020-04-29
D8NID-635 Add JS library to preprocess page: rewrite links

johangant:
Self referencing links in specific regions are now changed to span elements with specific classes

### DIFF
--- a/nidirect_common/js/link-manager.js
+++ b/nidirect_common/js/link-manager.js
@@ -1,0 +1,40 @@
+/**
+ * @file
+ * Defines Javascript behaviors for Link handling.
+ */
+(function ($) {
+
+  "use strict";
+
+  /**
+   * Changes any page links, within certain defined regions, from
+   * <a> elements to <span> elements if the current path matches the
+   * 'href' attribute.
+   *
+   * header.header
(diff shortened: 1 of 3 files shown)


nidirect_common/js/link-manager.js
```diff
@@ -0,0 +1,40 @@
...
+
+      $.each(elements, function(index, elementRef) {
+        let classes = ['active', 'link__self'];
+
+        if (elementRef === '#block-mainnavigation') {
+          classes.push('nav-link');
+        }
+
+        $(elementRef + ' a[href*="' + pathname + '"]').each(function() {
+          if ($(this).attr('href') === pathname) {
```
DuttonMa, line 32: Not sure that this will work for relative paths, might it be worth thinking of a regex here ?

johangant: Do you have an example for this? It's a somewhat crude matching process (contains) but `$(location).attr('pathname')` from http://example.com/article/a-title-value?arg1=1&arg2=2 yields `/article/a-title-value` so ought to suffice for the vast majority of cases.

I guess if we hit an edge case with something down the line, then it's a relatively simple thing to adjust.

PR #153, opened 2020-05-05
Preprocess media view listing

johangant:
- Add dependency to whatlinkshere module to fetch cross-referencing values
- Generate a new link element and replace the custom text field value in the view field object

### DIFF
--- a/nidirect_media/nidirect_media.module
+++ b/nidirect_media/nidirect_media.module
@@ -7,6 +7,8 @@
 
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Link;
+use Drupal\views\Render\ViewsRenderPipelineMarkup;
 
 /**
  * Implements hook_element_info_alter().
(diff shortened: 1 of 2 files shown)


nidirect_media/nidirect_media.module
```diff
@@ -136,3 +138,22 @@ function nidirect_media_form_alter(&$form, FormStateInterface $form_state, $form
...
+ *
+ * Injects the usage value of a media item into the table display.
+ */
+function nidirect_media_preprocess_views_view_field(&$variables) {
+  $view = $variables['view'];
+  $field_id = $variables['field']->field;
+
+  if ($view->id() == 'media' && $view->current_display == 'media_page_list' && $field_id == 'nothing') {
+    $mid = $variables['row']->mid;
+    $usage = \Drupal::service('whatlinkshere.linkmanager')->getReferenceContent($variables['row']->_entity, PHP_INT_MAX, 0);
```
johangant, line 153: `PHP_INT_MAX` used because this service function is sort of shoehorned/repurposed here, and I just needed a suitably large number.

nidirect_media/nidirect_media.module
```diff
@@ -136,3 +138,22 @@ function nidirect_media_form_alter(&$form, FormStateInterface $form_state, $form
...
+    $ref_count = $usage['total'];
```
DuttonMa, line 154: Would it be good to have some defensive coding here in case the service call hasn't worked and $usage is undefined ?

johangant: Normally I'd agree but on balance, there is minimal benefit because:

- the `getReferenceContent` function on the service always returns an array with the totals of the query even if they're empty.
- the `LinkManagerInterface` guarantees an array return type.

If there's something that breaches those contracts between service and interface then there's a much more fundamental problem or bug, which should be fixed.

Sort of feels like you _could_ put a `if empty()...` check but then becomes a case of how defensive you want to go. For me, an interface and sensible implementation is good enough (for now).

PR #154, opened 2020-05-06
D8 nid 658

omahm:
This is far from the ideal way to do this but other methods worked until the theme layer ripped out the markup to wrap the Featured content in a div to mimic the Homepage View.


omahm:
Pulling because it presents issues with LazyBuilder

PR #155, opened 2020-05-11
Rework link manager to use regex as matcher rather than simpler string matcher

johangant:


nidirect_common/js/link-manager.js
```diff
@@ -29,7 +32,14 @@
         }
 
         $(elementRef + ' a[href*="' + pathname + '"]').each(function() {
-          if ($(this).attr('href') === pathname) {
+          let href = $(this).attr('href');
+          // Prune away any query parameters, if they exist.
+          href = href.substr(0, (href.indexOf('?') < 0) ? href.length : href.indexOf('?'));
+
+          // Need a regex object to inject the pathname variable as a matching value.
+          let regex = new RegExp('^' + pathname + '$'.replace(/\//g, '\\/'), 'g');
```
johangant, line 40: Note the `^` and `$` delimiters to ensure we don't remove links from deeper content with a matching beginning segment of the URL path, eg, given a path to match of `/news`:

YES: `/news`
NO: `/news/something-else` or `/news/something-else?params=yes`

Without the trailing regex anchor you would accidentally clobber anything with a `/news` beginning in the URL.

PR #158, opened 2020-05-14
D8 nid 660

omahm:


omahm:
Could probably refactor the node/node_preview into a single function call but I've overspent the allowed time on this ticket.

PR #162, opened 2020-05-22
D8 nid 680

omahm:


nidirect_gp/js/gp-location-search.js
```diff
@@ -0,0 +1,54 @@
...
+(function ($, Drupal) {
+  Drupal.behaviors.gpLocationSearch = {
+    attach: (context) => {
+
+      // Callback function to initiate the GP location search.
+      function performLocationSearch(location) {
+        let lat = location.coords.latitude;
+        let lng = location.coords.longitude;
+
+        let url = '/services/gp-practices?lat=' + lat +'&lng=' + lng + '&proximity=10';
```
johangant, line 15: Is it possible to obtain/inject `nidirect_gp.proximity_max_distance` from nidirect_gp.services.yml to keep this in sync? Not likely to be a big deal but if that value is changed in future someone will need to remember to keep them both in sync

omahm: Yeah, good idea!

PR #170, opened 2020-06-02
D8 nid 700 gp search location theming

neilblair:
- Added class to div container for searching by location for JS/CSS hooks
- Added aria-live div inside the search by location div to use for displaying status updates when a location search is in progress
- Inserted a Drupal throbber (Drupal.theme.ajaxProgressIndicatorFullscreen()) into the aria-live region so we get an indication that something is happening when the "Use my location" button is clicked.  The throbber is removed if the location search fails.
- Added a timeout for the geolocation search
- Ensure focus goes to the normal searchbox following closure of the drupal dialog that appears when location search fails (no point focussing on the Use my location button).
- More tinkering with gp search titles ...

### DIFF
--- a/nidirect_gp/js/gp-location-search.js
+++ b/nidirect_gp/js/gp-location-search.js
@@ -7,6 +7,12 @@
   Drupal.behaviors.gpLocationSearch = {
     attach: (context) => {
 
+      let locationOptions = {
+        enableHighAccuracy: false,
+        timeout: 5000,
+        maximumAge: 0
+      };
+
       // Callback function to initiate the GP location search.
       function performLocationSearch(location) {
         let lat = location.coords.latitude;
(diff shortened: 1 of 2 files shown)


nidirect_gp/js/gp-location-search.js
```diff
@@ -18,8 +24,16 @@
...
 
       // Callback function for geolocation errors.
-      function locationError() {
-        const confirmationDialog = Drupal.dialog('<div>Unable to determine your location.</div>', {
+      function locationError(err) {
+
+        console.warn(`ERROR(${err.code}): ${err.message}`);
+        $('#find-by-location-status').html('');
+
+        let errmsg = '' +
```
omahm, line 32: Looking back at my code, ideally I should have passed it through Drupal's translation function and we should consider doing that here.

neilblair: Yep never thought of that - I've added it here and a couple of other spots.

PR #178, opened 2020-06-23
D8 nid 712

omahm:
Provides preview breadcrumbs for certain nodes. 
Also cleaned up any phpcs warnings for all breadcrumb classes.


omahm:
> Looks good to me. Nice to see preview and breadcrumbs are easier to work with in D8!

Breadcrumbs in D8 are a lot lot better!

PR #182, opened 2020-06-24
D8NID-751 ~ Hide link to toggle media weight display

omahm:


omahm:
Github seems to have screwed up this PR

PR #184, opened 2020-06-26
Take over some of linkit module's functionality

johangant:
- Required due to a lack of preprocess or easy override (via dependecy injection) of underlying service classes.

May be useful to split into an origins module if Unity requires it?

### DIFF
--- a/nidirect_common/src/LinkitResultManager.php
+++ b/nidirect_common/src/LinkitResultManager.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Drupal\nidirect_common;
+
+/**
+ * @file
+ * Contains \Drupal\linkit\ResultManager.
+ */
+
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Url;
+use Drupal\linkit\ProfileInterface;
+
+/**
+ * Result service to handle autocomplete matcher results.
+ */
+class LinkitResultManager {
+
+  /**
+   * Gets the results.
+   *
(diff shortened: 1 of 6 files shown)


nidirect_common/src/LinkitResultManager.php
```diff
@@ -0,0 +1,80 @@
...
+        'group' => t('E-mail'),
+      ];
+    }
+
+    // If there is still no matches, return a "no results" array.
+    if (empty($matches)) {
+      return [
+        [
+          'title' => t('No content results'),
+          'description' => t('There were no content results. Using the link provided as-is.'),
```
johangant, line 72: @neilblair would be good to get your input on the precise phrasing you'd like here.

Looks like this with these changes:

Screenshot 2020-06-26 at 17 54 19

nidirect_common/src/Routing/RouteSubscriber.php
```diff
@@ -0,0 +1,40 @@
...
+use Symfony\Component\Routing\RouteCollection;
+
+/**
+ * Listens to the dynamic route events.
+ */
+class RouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * Override from base class to set lighter
+   * event priority; execute after taxonomy_access_fix
```
johangant, line 16: Whoops. Copy/pasta error.

c595586b491836bb2135fb603c2f4e5997ce0e2f8e41c30a81af650f568eaff8

PR #192, opened 2020-07-03
Set fixed field labels for URLs only if they have a value

johangant:


nidirect_gp/nidirect_gp.module
```diff
@@ -146,6 +146,33 @@ function nidirect_gp_gp_practice_form_validate(&$form, FormStateInterface $form_
...
+
+    if (empty(reset($field['uri']))) {
+      // Skip items with blank URI values.
+      continue;
+    }
+
+    // If we have a URI value then we want to use fixed labels for the link
+    // text on these fields.
+    if ($field_id == 'field_gp_prescriptions') {
+      $field[0]['title'] = t('Online appointments');
```
omahm, line 165: Looks like you have a Copy 'n' Pasta error, should be 'Online prescriptions'

johangant: D'oh... clipboard fail again!

PR #198, opened 2020-07-24
D8 nid 512 c

DuttonMa:
Create NIDirect specific users for Nightwatch testing in hook_install - and delete them in hook_uninstall

### DIFF
--- a/nidirect_common/nidirect_common.install
+++ b/nidirect_common/nidirect_common.install
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * @file
+ * Install, update and uninstall functions for the NIDirect common module.
+ */
+
+use Drupal\user\Entity\User;
+
+/**
+ * Implements hook_install().
+ *
+ * Perform actions to set up the site for this module.
+ */
+function origins_workflow_install() {
+  // Create some test users for the Nightwatch tests that will come along later.
+  $name_list = [
+    '_author' => 'author_user',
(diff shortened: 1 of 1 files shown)


nidirect_common/nidirect_common.install
```diff
@@ -0,0 +1,78 @@
...
+    '_hc_super' => 'health_condition_supervisor_user',
+    '_driving_super' => 'driving_instructor_supervisor_user',
+  ];
+  foreach ($name_list as $name => $role) {
+    // Add prefix from environment var.
+    $prefix = getenv('NW_TEST_USER_PREFIX');
+    $password = getenv('TEST_PASS');
+    // If prefix not set, do not create users.
+    if (empty($prefix) || empty($password)) {
+      \Drupal::logger('nidirect_common')->notice(
```
omahm, line 31: I'd be tempted to use 'error' vs 'notice'

johangant:
Looks decent to me. These accounts might be a little redundant in future, depending on where we run our tests. For example: I might choose to create a user account/remove it as part of a CI job - which would probably be preferable to having fixed test accounts when the site goes live - but that's not necessarily shared with Lando so probably needs more thought and refinement.

nidirect_common/nidirect_common.install
DuttonMa: Good spot, thanks

PR #203, opened 2020-07-30
D8NID-812 Only swap render array on news FCL node, not homepage.

johangant:
Also adds a regression test to detect anything like it in future.

### DIFF
--- a/nidirect_news/nidirect_news.module
+++ b/nidirect_news/nidirect_news.module
@@ -35,10 +35,21 @@ function nidirect_news_preprocess_field(array &$variables) {
  * Implements hook_preprocess_node().
  */
 function nidirect_news_preprocess_node(array &$variables) {
-  if ($variables['node']->getType() == 'featured_content_list' && $variables['view_mode'] == 'full') {
-    // Replace the node view with a rendered view display, using the node items as arguments.
-    // This mirrors the display you would see on the embedded views display.
-    $variables['content'] = \Drupal::service('nidirect_news.news')->getFeaturedNews();
+  if ($variables['node']->getType() != 'featured_content_list') {
+    return;
+  }
+
+  if ($variables['view_mode'] == 'full' && !empty($variables['node']->field_tags->target_id)) {
+    $news_tag = \Drupal::entityTypeManager()
+      ->getStorage('taxonomy_term')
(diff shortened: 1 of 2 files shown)


nidirect_news/tests/src/Nightwatch/Tests/featuredContentListViewSwitch.js
```diff
@@ -0,0 +1,37 @@
+let newsContentTitles = [];
+let homepageFeaturedTitles = [];
+
+module.exports = {
+  '@tags': ['regression'],
+
+  before: browser => {
+    // Gather the titles of the current 'latest/featured news' content.
+    browser.drupalRelativeURL('/news');
+    browser.elements('css selector', '.card .card__title', function (result) {
```
johangant, line 10: I cannot tell you how long this took to piece together. Between finding a way to handle multiple values from the CSS selector to storing them in an accessible variable for use later on... mind blown.

nidirect_news/tests/src/Nightwatch/Tests/featuredContentListViewSwitch.js
```diff
@@ -0,0 +1,37 @@
+let newsContentTitles = [];
+let homepageFeaturedTitles = [];
+
+module.exports = {
+  '@tags': ['regression'],
```
johangant, line 5: Felt like a useful, general tag to keep to ensure we can easily check for all previously detected regressions.

nidirect_news/nidirect_news.module
```diff
@@ -35,10 +35,21 @@ function nidirect_news_preprocess_field(array &$variables) {
...
+    return;
+  }
+
+  if ($variables['view_mode'] == 'full' && !empty($variables['node']->field_tags->target_id)) {
+    $news_tag = \Drupal::entityTypeManager()
+      ->getStorage('taxonomy_term')
+      ->loadByProperties(['name' => 'News']);
+    $news_tag = reset($news_tag);
+
+    if ($variables['node']->field_tags->target_id == $news_tag->id()) {
```
johangant, line 48: Crucially, this was missing before. Without it, it swaps the homepage FCL render array around as well which is not what we want!

PR #209, opened 2020-08-14
D8 nid 831

omahm:


nidirect_landing_pages/nidirect_landing_pages.module
```diff
@@ -182,7 +182,7 @@ function nidirect_landing_pages_form_alter(&$form, FormStateInterface $form_stat
       $msg = "Note that the theme/subtheme that you choose here will be overridden by this landing page. <br/>";
       $msg .= "As an example, if you select 'Motoring / Road Safety' here then when you visit the page at ";
       $msg .= "/information-and-services/motoring there will be a link to this page labelled 'Road Safety'.";
-      $form['field_subtheme']['widget']['#description'] = t($msg);
+      $form['field_subtheme']['widget']['#description'] = $msg;
```
johangant, line 185: Did `t()` not work for the description text here?

nidirect_landing_pages/nidirect_landing_pages.routing.yml
```diff
@@ -0,0 +1,8 @@
+
+nidirect_landing_pages.landing_pages_choose_block_controller_build:
+  path: '/nidirect_landing_pages/chooseblock'
+  defaults:
+    _controller: '\Drupal\nidirect_landing_pages\Controller\LandingPagesChooseBlockController::build'
+    _title: 'build'
+  requirements:
+    _permission: 'access content'
```
johangant, line 8: Does this mean anonymous users can also call this controller method? Can we set it to be something like 'edit landing pages' which editors are more likely to have?

nidirect_landing_pages/nidirect_landing_pages.module
omahm: Code standards will whine about it.

nidirect_landing_pages/nidirect_landing_pages.routing.yml
omahm: Doh, my bad

PR #213, opened 2020-08-20
D8 nid 820

DuttonMa:
Correction to breadcrumb trail on landing pages


johangant (commented):
So if I understand this, we're removing 'Home' from Landing page nodes? If we extend `NodeThemesBreadcrumb::applies()` it might allow you to re-use and adjust rather than duplicate?

If not, this looks okay. I reckon by the end of the project we'll need to review/refactor a good amount of the breadcrumb handlers.

PR #216, opened 2020-08-24
D8NID-719 Change label on shs default value to match regular dropdown options

johangant:
Well this threw me down the garden path. There are a range of hooks allowing you to adjust the SHS widget's options and, at face value, it seems to only show get methods on the widget object making it hard to set a new value in the way it would expect.

However, because the widget object type is a translated markup object, I'm proposing to cheat a little here and replace the value with a translated text value. As far as I can tell, it seems like a safe operation.


omahm (approved):
Welcome to the joys of SHS module ;)

PR #217, opened 2020-08-26
D8NID-845 regression test for access issue on taxonomy term

johangant:
Checks that we have the 'Help' item in the nav menu and that we aren't looking at an access denied page.


omahm (approved):
LG

PR #220, opened 2020-08-28
D8 nid 830

omahm:
PR for review. Issue whereby the landing page blocks are generated twice, initially as part of the import process and again on node save.


nidirect_landing_pages/modules/nidirect_campaign_importer/nidirect_campaign_importer.info.yml
```diff
@@ -0,0 +1,8 @@
+name: 'NIDirect Campaign Importer'
+type: module
+description: 'Import Drupal 7 campaign pages as landing pages.'
+core_version_requirement: ^8.8 || ^9
```
johangant, line 4: Ooo, nice. Do we want to work this into our other custom modules too?

omahm: Yup, I've started changing them, will throw up a PR soon

johangant (approved):
Looks great to me. I think it'll be a big help to jumpstart the content creation process for layout builder powered landing pages

DuttonMa:
Great work, I like the way that you have broken the code down (and also some nice use of XPath in there too).
Just a couple of small queries really.

PR #226, opened 2020-09-08
D8 nid 866

omahm:
Initial port of custom suggestion manager for the version 6 update of the LinkIt module.


johangant (approved):
Nice to see a lot of the fudgy code settling into something more cohesive with a later module release.

LGTM 👍

PR #234, opened 2020-09-23
Fix handling of pause when populating title field

johangant:
The unique title field check has an AJAX call which introduces a slight pause when moving away from the title field. We need to both trigger a key press to activate the JS event and then briefly pause to let it do its thing before moving onto populating the rest of the form.


omahm (approved):
LG

PR #238, opened 2020-09-24
Health conditions search prompt

neilblair:
Adds JS/CSS to the health conditions module to display the "search conditions prompt" when viewing health conditions on a narrow screen.  

The prompt appears at the bottom of the screen after scrolling down a bit.  Clicking the prompt skips you down to the search/a-z widget at the bottom of the page and sets focus on the search input.

Screenshot 2020-09-24 at 13 42 06


johangant (approved):
Looks good 👍

PR #245, opened 2020-09-25
D8 nid 354

DuttonMa:
Custom code to overcome an issue in Drupal 8 core views that prevents sorting by 'link' fields.

Relates to https://github.com/dof-dss/nidirect-drupal/pull/455

### DIFF
--- a/nidirect_gp/nidirect_gp.module
+++ b/nidirect_gp/nidirect_gp.module
@@ -252,22 +252,36 @@ function _build_gp_practice_title($practiceName = '', $surgeryName = '') {
  * Implements hook_views_query_alter().
  */
 function nidirect_gp_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
+  // Alter the view code for GP Practice Lists if being sorted on the
+  // 'Online appointments' or 'Online prescriptions' fields.
+  // This is necessary because Drupal 8 Views does not seem to handle the case
+  // of sorting on a 'link' field as it does not build the join correctly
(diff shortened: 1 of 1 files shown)


nidirect_gp/nidirect_gp.module
```diff
@@ -246,3 +248,62 @@ function _build_gp_practice_title($practiceName = '', $surgeryName = '') {
...
+        // Decide whether we are sorting by the 'Online appointments' column or
+        // by 'Online precriptions'.
+        $join_table = NULL;
+        $view_input = $view->getExposedInput();
+        if (isset($view_input) && isset($view_input['order'])) {
+          if ($view_input['order'] == 'nothing') {
+            // We are sorting by 'Online appointments'.
+            $join_table = 'node__field_gp_appointments';
+          }
+          elseif ($view_input['order'] == 'nothing_1') {
```
johangant, line 276: These unknown and nothing values that views gives really aren't terribly helpful for us, but it's done it for years so.... :)

johangant (approved):
Pretty dense in terms of conditionals, but I reckon it'll get the outcome we need. Would be good to see a nightwatch test here to alert if this sorting pattern breaks in future

omahm (approved):

nidirect_gp/nidirect_gp.module
DuttonMa: Yes, it doesn't seem possible to change those - I have tried !

PR #250, opened 2020-10-08
D8 nid 842

DuttonMa:
Alter the system performance admin screen so that a 'Browser and proxy cache maximum age' of 1 year may be selected.
Also add an informative suffix to make it clear that this has been done.

Related to https://github.com/dof-dss/nidirect-drupal/pull/465


omahm (approved):
Nice one

PR #252, opened 2020-10-13
Instruct clientside_validation module not to operate on non-webform forms

johangant:
A little more dancing around this module. It's not practical to target a multitude of form ids and routes, so inverting this and only letting it act on webform forms seems more pragmatic.

It seems to still need two hooks at the very least, but the `#after_build` removal is something from before that can probably stay for now even though it might not be needed with the page_attachments_alter and this hook in place,.


DuttonMa (approved):
Looks good

PR #255, opened 2020-10-19
D8NID-822 Wrap featured news embed into a block

johangant:
Re-use as much of the broader news views display approach as possible. Essentially use views as a rendering method for news node ids passed in as contextual arguments. Featured news display differs from the 'latest news' one on the landing page because it uses fields (like the block display) for the sidebar


omahm (approved):
LGTM

PR #258, opened 2020-10-23
D8NID-912 improve search elevation handling

johangant:
Gets the index hash from config rather than having it hard coded. Streamlines some of the query preprocessing, too.

Snuck in a small tidyup in contacts module to remove two unused variables.


omahm (approved):
Nice one,  good comprehensive documentation too!

PR #260, opened 2020-10-27
D8NID-906

johangant:
Add a service for getting at health condition symptom values and preprocess the JSON-LD graph to add extra information into the array that would otherwise be harder to manage with tokens alone


DuttonMa (approved):
Nice work

PR #262, opened 2020-11-02
Switch regression test to use xpath for some elements and reinstate hook to alter page attachments.

johangant:
Turns out clientside_validation cannot be turned off with just the hook it proposes you use.


omahm (approved):
Love me some xpath 👍

PR #268, opened 2020-11-16
D8NID-951

johangant:
Preprocesses the related content view so that users who have high editorial access permissions do see content that would ordinarily be hidden in the sidebar by the 'Hidden' flag being set.


omahm (approved):
LG, nice comment too :)

PR #269, opened 2020-11-17
Switch off lazy builder placeholders on some blocks

johangant:
Causes issues when checking for region/variable contents in page templates, because we don't know if the render array is empty because it is build/rendered afterwards by the lazy builder service


omahm (approved):
Nice one.

PR #272, opened 2020-11-24
Fix HC alt add/edit on HC nodes.

johangant:
Hide on HC creation form as (for now) we rely on the node existing before attaching HC alt nodes that are created.


johangant:
Superseded by changes and fixes in https://digitaldevelopment.atlassian.net/browse/D8NID-981

PR #277, opened 2020-11-27
D8NID-969

johangant:


nidirect_news/src/Plugin/Block/FeaturedNewsBlock.php
```diff
@@ -18,7 +20,9 @@ class FeaturedNewsBlock extends BlockBase {
    * {@inheritdoc}
    */
   public function build() {
-    $content['featured_news'] = \Drupal::service('nidirect_news.news')->getNewsEmbed('featured_news');
+    $view = \Drupal::service('nidirect_news.news')->getLatestNewsView();
```
DuttonMa, line 23: Would it be better to use DI here ?

DuttonMa (approved):
Nice work, shame it had to be done & re-done !
Just a query about Dependency Injection

nidirect_news/src/Plugin/Block/FeaturedNewsBlock.php
johangant: Yep, in hindsight it couldn't hurt. Means a bit of extra boilerplate to go in but shouldn't take too long.

omahm (approved):
Looks good and as per Martin's comments, would be better to inject those services.

nidirect_news/src/Plugin/Block/FeaturedNewsBlock.php
johangant: Pushed an update this morning to sort this.

PR #280, opened 2020-11-30
Hide 'sticky at top of lists' option on nodes except news and contacts.

neilblair:
Make the 'sticky at top of lists' option available on contacts.

### DIFF
--- a/nidirect_common/inc/form_alter.inc
+++ b/nidirect_common/inc/form_alter.inc
@@ -128,7 +128,7 @@ function nidirect_common_form_node_form_alter(&$form, FormStateInterface $form_s
   $form['promote']['#access'] = FALSE;
 
   // Hide 'Sticky at top of lists' option, except for news and contact nodes.
-  if (!preg_match('/^node_news/', $form['#form_id']) && !preg_match('/^node_contact/', $form['#form_id'])) {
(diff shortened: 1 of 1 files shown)


nidirect_common/inc/form_alter.inc
```diff
@@ -127,8 +127,8 @@ function nidirect_common_form_node_form_alter(&$form, FormStateInterface $form_s
   // Hide 'promote to front-page' option.
   $form['promote']['#access'] = FALSE;
 
-  // Hide 'Sticky at top of lists' option, unless it's a news node.
-  if (!preg_match('/^node_news/', $form['#form_id'])) {
+  // Hide 'Sticky at top of lists' option, except for news and contact nodes.
+  if (!preg_match('/^node_news/', $form['#form_id']) && !preg_match('/^node_contact/', $form['#form_id'])) {
```
johangant, line 131: For a little more conciseness, how about `if (!preg_match('/^node_(news|contact)/', $form['#form_id'])) {` ?

johangant (approved):
All good. Just the one suggestion for making the if statement a little more concise. Whether it's actually clearer or not I guess depends on the reader's eye for regex syntax :)

nidirect_common/inc/form_alter.inc
neilblair: Yes - nice one

PR #290, opened 2020-12-07
D8 nid 995

omahm:
- Moved payments processing from REST resource to it's own service.
- Copied payments and stations entity config to module for import during test.
- Created basic kernel test setup.


johangant (approved):
Looks good. Nice to expose the CWP service as a Drupal service for max re-use between tests and other code 👍

PR #293, opened 2020-12-08
D8 nid 1006

neilblair:
Adds link to sort contacts search results by relevance or alphabetically.
Fixes a bug in link-manager.js causing link processing loop to stop when link with data-self-ref attribute is encountered.


omahm (approved):
LG 👍

PR #297, opened 2020-12-15
D8NID-1017 ~ Update GP routes and menus

omahm:


nidirect_gp/src/Entity/Gp.php
```diff
@@ -59,7 +59,7 @@
  *   },
  *   links = {
  *     "canonical" = "/gp/{gp}",
- *     "add-form" = "/gp/add",
+ *     "add-form" = "/node/add/gp",
```
johangant, line 62: How does this work alongside GP Practice nodes? GPs are custom entities rather than node entities so I wouldn't expect that route to work

omahm: It didn't seem to have any impact.

PR #300, opened 2020-12-17
Add links to book parent(s) into the breadcrumb if node is in a book.

neilblair:


omahm (approved):
LG

nidirect_breadcrumbs/src/NodeThemesBreadcrumb.php
```diff
@@ -133,6 +133,26 @@ public function build(RouteMatchInterface $route_match) {
...
+    if (!empty($node->book)) {
+
+      // Determine depth of node in the book.
+      $book_depth = $node->book['depth'];
+
+      // Create links to parent nodes in the depths above - if published.
+      $i = 1;
+      while ($i < $book_depth) {
+        $p = 'p' . $i++;
+        $nid_published = \Drupal::entityQuery('node')
```
johangant, line 146: As a rule of thumb, when you have a custom class using direct service calls (`\Drupal::serviceName` or `Drupal::service('service_id')`) you will want to pass these in using dependency injection. That's not normally as hard as it sounds... usually a constructor, a static create function and some internal variables to represent the service objects.

In this instance, I'll propose progressive optimisation. So merge this (because it works and is safe) but make a note or JIRA ticket to swing back and introduce injected services to this class.

johangant (approved):
Looks good, nice use of internal services to get the job done. I left a comment about passing service objects in via dependency injection. I don't think it's a blocking issue but would be good to introduce that when possible as a matter of best practice. There'll be plenty of examples in the codebase by now but happy to support with this if needed.

neilblair:
> Looks good, nice use of internal services to get the job done. I left a comment about passing service objects in via dependency injection. I don't think it's a blocking issue but would be good to introduce that when possible as a matter of best practice. There'll be plenty of examples in the codebase by now but happy to support with this if needed.

@johangant could you run your eye over this again.  I had a go at proper dependency injection - took a while to get it right. I got stuck on an error

```
ArgumentCountError: Too few arguments to function Drupal\nidirect_breadcrumbs\NodeThemesBreadcrumb::__construct(), 1 passed in /app/drupal8/web/core/lib/Drupal/Component/DependencyInjection/Container.php on line  
                                           269 and exactly 2 expected in Drupal\nidirect_breadcrumbs\NodeThemesBreadcrumb->__construct() (line 60 of /app/drupal8/web/modules/custom/nidirect_breadcrumbs/src/NodeThemesBreadcrumb.php)
```

Eventually I worked out I had to add the missing argument into nidirect_breadcrumbs/nidirect_breadcrumbs.services.yml - but the error message gave no clue at all!

johangant:
> > Looks good, nice use of internal services to get the job done. I left a comment about passing service objects in via dependency injection. I don't think it's a blocking issue but would be good to introduce that when possible as a matter of best practice. There'll be plenty of examples in the codebase by now but happy to support with this if needed.
> 
> @johangant could you run your eye over this again. I had a go at proper dependency injection - took a while to get it right. I got stuck on an error
> 
> ```
> ArgumentCountError: Too few arguments to function Drupal\nidirect_breadcrumbs\NodeThemesBreadcrumb::__construct(), 1 passed in /app/drupal8/web/core/lib/Drupal/Component/DependencyInjection/Container.php on line  
>                                            269 and exactly 2 expected in Drupal\nidirect_breadcrumbs\NodeThemesBreadcrumb->__construct() (line 60 of /app/drupal8/web/modules/custom/nidirect_breadcrumbs/src/NodeThemesBreadcrumb.php)
> ```
> 
> Eventually I worked out I had to add the missing argument into nidirect_breadcrumbs/nidirect_breadcrumbs.services.yml - but the error message gave no clue at all!

Excellent stuff... DI's a bit weird to get used to at first but between the usual constructor + container create methods it's easy to see the flow of it. Having a look in `core.services.yml` is also useful to track down core service ids... there's a Drupal console command too that helps you list them out across the entire site.

PR #301, opened 2020-12-17
Keep HC alt title node status in sync with parent HC node status

johangant:
Easy way to ensure consistency between db based queries and Solr queries (all use entity status as a filter). Also allows us to avoid writing bespoke Solr processor plugins to do the same thing in a more convoluted manner.


omahm (approved):
LG

PR #305, opened 2021-01-04
D8NID-1028 External link node titles and urls come from the URL field value in articles_by_term listings

johangant:


omahm (approved):
Nice one

PR #306, opened 2021-01-07
Creating a new HC node should also sync the HCA node's publish status.

johangant:
Additionally, preprocess the url value in search result view mode so we don't point at the HCA node itself, rather the parent condition node.


omahm (approved):
LGTM

PR #307, opened 2021-01-08
Add breadcrumb builder class for stealing (some) control from core taxo module

johangant:
Intends to provide a more consistent UX when adding new terms to a vocab (less crossover with core taxonomy module's breadcrumb trail)

### DIFF
--- a/nidirect_taxonomy_navigator/src/TaxonomyAdminBreadcrumb.php
+++ b/nidirect_taxonomy_navigator/src/TaxonomyAdminBreadcrumb.php
@@ -1,0 +1,161 @@
+<?php
+
+namespace Drupal\nidirect_taxonomy_navigator;
+
+use Drupal\Component\Utility\Unicode;
+use Drupal\Core\Access\AccessManagerInterface;
+use Drupal\Core\Breadcrumb\Breadcrumb;
+use Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Controller\TitleResolverInterface;
+use Drupal\Core\Link;
+use Drupal\Core\Path\CurrentPathStack;
+use Drupal\Core\Path\PathMatcherInterface;
+use Drupal\Core\PathProcessor\InboundPathProcessorInterface;
+use Drupal\Core\Routing\RequestContext;
+use Drupal\Core\Routing\RouteMatch;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Url;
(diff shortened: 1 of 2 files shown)


nidirect_taxonomy_navigator/src/TaxonomyAdminBreadcrumb.php
```diff
@@ -0,0 +1,161 @@
...
+      $match = TRUE;
+    }
+
+    return $match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(RouteMatchInterface $route_match) {
```
johangant, line 108: A good chunk of this is duplicated from Core's fallback class... because we're doing more or less the exact same thing. Key differences are:

- In `applies()` we're checking the route against the service parameters, not just kicking in by default.
- In `build()` we're tweaking the path reference along the likes of `s/taxonomy\/manage/taxonomy_navigator/g`, and letting the remaining access checks and link generation happen as it should. It would have been lovely to be able to alter the path components before passing this back to the parent class for building, but because there's no 'hook' or event here to do that we end up having to take most of the original implementation and include it as part of the overridden method. Not as DRY as I'd like, but necessary.

nidirect_taxonomy_navigator/src/TaxonomyAdminBreadcrumb.php
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    while (count($path_elements) > 1) {
+      array_pop($path_elements);
+
+      // Copy the path elements for up-casting.
+      $route_request = $this->getRequestForPath('/' . implode('/', $path_elements), $exclude);
+
+      if ($route_request) {
+        $route_match = RouteMatch::createFromRequest($route_request);
```
omahm, line 138: Possibly inject the current route path?

omahm (commented):
Looks good, just one possible opportunity to use DI

nidirect_taxonomy_navigator/src/TaxonomyAdminBreadcrumb.php
johangant: Yep, agreed! Funnily enough this was lifted from core... unclear why it wasn't passed in there too.

johangant: Ah, it's a global static function... that'll be why then! https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Routing%21RouteMatch.php/function/RouteMatch%3A%3AcreateFromRequest/8.2.x

omahm: Ahh that explains it then, good find.

PR #308, opened 2021-01-11
D8 nid 1019

omahm:


nidirect_related_content/src/RelatedContentManager.php
```diff
@@ -0,0 +1,314 @@
...
+        }
+      }
+      else {
+        return $this;
+      }
+    }
+
+    $this->setTerms($term_ids);
+
+    if ($content === 'themes') {
```
johangant, line 115: Worth referencing the class constants here instead of literal string values? Eg: `$this-> CONTENT_THEMES` instead of `themes`

johangant (approved):
Looks great to me, a good healthy amount of refactoring to consolidate a few streams of custom code that have evolved over time.

A small comment about use of the in-class constants but otherwise looks good to go!

nidirect_related_content/src/RelatedContentManager.php
omahm: Doh,  very true.

DuttonMa (changes requested):
Looks like great work to me.
Just a thought - It looks to me like you have removed the only places that _render_articles_by_term() and _render_site_subtopics() were called from - so you should presumably delete those functions to avoid confusion (looks like you have moved the code elsewhere)

omahm:
> Looks like great work to me.
> Just a thought - It looks to me like you have removed the only places that _render_articles_by_term() and _render_site_subtopics() were called from - so you should presumably delete those functions to avoid confusion (looks like you have moved the code elsewhere)

Yup, good point.. cleanup after myself :D

PR #309, opened 2021-01-12
Turn off JS from link-manager.js on facet links

johangant:
Without this, the facet link is stripped leaving text alone which is not useful to the end-user


omahm (approved):
👍

PR #312, opened 2021-01-15
D8NID-1047 - tweaks to add term form for site themes:

neilblair:
- open relations section by default
- enable chosen on the parent select element
- allow only one parent to be selected


omahm (approved):
LGTM

PR #314, opened 2021-01-20
hide article teaser listing fields on node edit form

DuttonMa:
Do not show the 'manually control listing' checkbox or the listing fields when editing or creating a landing page node. This functionality is now available in layout builder only.


johangant:
Can we control this with the form display config at all instead of using form API?

DuttonMa:
> Can we control this with the form display config at all instead of using form API?

Yes, we could do this but I have a feeling that I did this before and it was then 'un-done' ! Personally, it feels more robust having it in code.

johangant:
> > Can we control this with the form display config at all instead of using form API?
> 
> Yes, we could do this but I have a feeling that I did this before and it was then 'un-done' ! Personally, it feels more robust having it in code.

Okay, no problem. Let's go with this for now and I'm sure it can be revised in future, if needed.

johangant (approved):

PR #323, opened 2021-02-01
D8NID 1060

johangant:
- Detect and prune search strings with absolute paths in them
- Inject request stack service


DuttonMa (approved):
Looks good

PR #330, opened 2021-02-09
Driving instructors: replace categories exposed filter 

neilblair:
with radios compatible with facets and add help text to top of the form (hidden when results are shown)

### DIFF
--- a/nidirect_driving_instructors/nidirect_driving_instructors.module
+++ b/nidirect_driving_instructors/nidirect_driving_instructors.module
@@ -251,4 +251,3 @@ function nidirect_driving_instructors_preprocess_facets_item_list(array &$variab
     $facet['value']['#attributes']['data-self-ref'] = 'false';
   }
 }
-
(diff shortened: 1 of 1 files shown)


nidirect_driving_instructors/nidirect_driving_instructors.module
```diff
@@ -38,29 +39,56 @@ function nidirect_driving_instructors_form_alter(&$form, &$form_state, &$form_id
...
+    // See driving instructors facet source configuration (/admin/config/search/facets)
+    // for the correct filter key to use.
+    $form['f[0]'] = [
+      '#id' => Html::getUniqueId('edit-categories'),
+      '#type' => 'radios',
+      '#title' => t('Filter by type of tuition'),
+      '#title_display' => 'above',
+      '#options' => [
+        'categories:all' => t('All'),
+        'categories:1160'  => t('Driving instructors'),
```
johangant, line 72: There's a potential brittle area here with the term ids baked into the form structure values. I figure these are unlikely to change much as much of the IA is established, but ideally, these term ids would come from using `\Drupal::entityTypeManager()->getStorage('taxonomy_term')->loadByProperties(['name' => 'Driving instructors']);` .... then extracting the term ids from the object that's returned.

johangant (changes requested):
Looks great, but the term ids for the form options should really come from a taxonomy term query to avoid future breakages in case term ids change.

nidirect_driving_instructors/nidirect_driving_instructors.module
neilblair: Thanks Johan - could you take another look at this please?

johangant (approved):
Much better! Good to go 👍

PR #331, opened 2021-02-11
D8NID-1088

DuttonMa:
Added JQuery to deal with more than one webform on a page


omahm (approved):

johangant (approved):
LGTM. I can see that webform module itself uses a similar process to handle other clashes. A pity it can't somehow dovetail in with https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AgetUniqueId/8.6.x but that's hopefully something for a future core release :)

PR #338, opened 2021-02-18
D8 nid 1094 cold weather payments form

neilblair:
Enable the form to accept a full NI postcode plus changes to theme it up and add styled error message.

Screenshot 2021-02-18 at 16 00 42

### DIFF
--- a/nidirect_cold_weather_payments/src/Form/ColdWeatherPaymentCheckerForm.php
+++ b/nidirect_cold_weather_payments/src/Form/ColdWeatherPaymentCheckerForm.php
@@ -87,6 +87,10 @@ class ColdWeatherPaymentCheckerForm extends FormBase {
       '#weight' => '0',
       '#title' => t('Enter your Northern Ireland postcode'),
       '#title_display' => 'before',
+      '#attributes' => [
+        'autocomplete' => 'postal-code',
+      ],
+      '#default_value' => $form_state->getValue('postcode', ''),
     ];
 
     $form['submit'] = [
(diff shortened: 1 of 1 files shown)


nidirect_cold_weather_payments/src/Form/ColdWeatherPaymentCheckerForm.php
```diff
@@ -224,4 +252,28 @@ private function cwpLookup($postcode) {
...
+   * Get postcode district from a Northern Ireland postcode.
+   *
+   * For Northern Ireland postcodes, the postcode district is always 1 or 2 digits following the postcode area 'BT'.
+   */
+  private function cwpGetPostcodeDistrict(string $postcode) {
+    $postcode_district = NULL;
+
+    // If postcode is a full NI postcode, or just the first part (outward code - e.g. BT1) ...
+    if (preg_match('/^[bB][tT][0-9]{1,2}( ?[0-9][a-zA-Z]{2})?$/', $postcode)) {
+      if (strlen($postcode) > 4) {
```
johangant, line 265: I'd recommend you trim any spaces in `$postcode` here as a padding space between the two halves can affect your string length check and potentially some of the substring trim points could be offset by 1.

johangant (changes requested):
Looks good. A space removal in `cwpGetPostcodeDistrict` would help avoid any potential offset issues with trimming the string to calculate the district code.

nidirect_cold_weather_payments/src/Form/ColdWeatherPaymentCheckerForm.php
omahm: I'd also consider lowercasing in the first instance so you can reduce the regex expression.

neilblair: @johangant , @omahm  - see latest commit.

johangant (approved):
👍  nice one

PR #342, opened 2021-02-26
D8 nid 995

omahm:
- Also tidied up some other areas that broke code standards


DuttonMa (approved):
Nice work, very readable

PR #347, opened 2021-03-05
D8 nid 1109

DuttonMa:
Changes for new version 8.x-3.16 of the geocoder module:
- Now pass in a provider to geocoder rather than just the provider name e.g. 'googlemaps'
- Code added to overwrite the Google maps 'placeholder' api key that is stored in config with the real api key from the environment variable
- Only send full postcodes to geocoder, as it does not work with short ones like 'BT5'

Related to https://github.com/dof-dss/nidirect-drupal/pull/628


nidirect_gp/src/Controller/GpSearchController.php
```diff
@@ -233,6 +261,23 @@ public function handleSearchRequest() {
 
   }
 
+  /**
+   * Update the Google maps api key if necessary.
+   */
+  private function updateApiKey() {
```
omahm, line 267: I'm maybe consider having this outside the GPsearch controller so we can use it across the site.

nidirect_gp/nidirect_gp.services.yml
```diff
@@ -9,4 +9,4 @@ parameters:
 
   # Regex sourced/tweaked from https://andrewwburns.com/2018/04/10/uk-postcode-validation-regex/.
   # Accepts formats such as: SW1, SW1A, SW1A1AA, SW1A 1AA
-  postcode_matches.uk: '^(([A-Z][0-9]{1,2})|(([A-Z][A-HJ-Y][0-9]{1,2})|(([A-Z][0-9][A-Z])|([A-Z][A-HJ-Y][0-9]?[A-Z]))))((\s?)([0-9][A-Z]{2})?)$'
+  postcode_matches.uk: '^(([A-Z][0-9]{1,2})|(([A-Z][A-HJ-Y][0-9]{1,2})|(([A-Z][0-9][A-Z])|([A-Z][A-HJ-Y][0-9]?[A-Z]))))((\s+)([0-9][A-Z]{2})?)$'
```
DuttonMa, line 12: We need to enforce a space here between the two parts of the postcode.
This is because the geocoder won't work with partial postcodes like 'BT5', a full postcode like 'BT5 8XX' must be supplied.

nidirect_gp/src/Controller/GpSearchController.php
```diff
@@ -233,6 +261,23 @@ public function handleSearchRequest() {
...
+    // Ensure that the geocoder provider api key is correct (as the api key
+    // cannot be held in config, it may be necessary to update it here with
+    // the api key that is held in the environment variable).
+    $googlemap_config = $this->configFactory->getEditable('geocoder.geocoder_provider.googlemaps')->get('configuration');
+    if (!empty($googlemap_config) &&
+      isset($googlemap_config['apiKey']) &&
+      ($googlemap_config['apiKey'] != getenv('GOOGLE_MAP_API_SERVER_KEY'))) {
+      // Overwrite the google map api key.
+      $googlemap_config['apiKey'] = getenv('GOOGLE_MAP_API_SERVER_KEY');
+      $this->configFactory->getEditable('geocoder.geocoder_provider.googlemaps')->set('configuration', $googlemap_config)->save();
```
DuttonMa, line 277: This is not ideal but is necessary to avoid holding an api key in config.
It has been optimised as much as possible, so only the first GP search will actually do this update after each config import.

omahm (commented):
Good job!   I'd maybe spin the API key assignment out to a simple class that we can use across the site?

nidirect_gp/nidirect_gp.services.yml
johangant: A postcode prefix definitely worked in the past, see https://github.com/dof-dss/nidirect-site-modules/blob/652d19188078595fb83ae8233653bc7eb770a631/nidirect_search/tests/src/Nightwatch/Tests/gpSearchTests.js#L31

Could be geocoder config that needs adjusting to accommodate a certain level of specificity?

DuttonMa: Good point, I have now added a region setting to the geocoding provider and it works with partial postcodes

nidirect_gp/src/Controller/GpSearchController.php
DuttonMa: Good thought, have done that now (see new class UpdateConfigFromEnvironment)

nidirect_gp/src/Controller/GpSearchController.php
```diff
@@ -91,14 +99,17 @@ class GpSearchController extends ControllerBase {
...
+   *   Entity type manager.
    */
   public function __construct(
     RequestStack $request_stack,
     PostcodeExtractor $postcode_extractor,
     GeocoderInterface $geocoder,
     int $proximity_max_distance,
     string $geocoding_service_id,
-    FormBuilder $form_builder
+    FormBuilder $form_builder,
```
johangant, line 111: FormBuilderInterface would be ideal here as it means you can pass in a test class that implements that, if needed.

DuttonMa: OK, have done that now

johangant (approved):
All good now. Cheers 👍

PR #352, opened 2021-03-16
Use block config instead of taxonomy tags for referenced FCL nodes

johangant:


johangant:
Sigh... TFW when you realise your git branch references the wrong ticket id right at the very end 🤦

PR #353, opened 2021-03-16
Use block config instead of taxonomy tags for referenced FCL nodes

johangant:
There's a fair bit going on there but in summary:

- Adds block form + submit handlers to define options for FCL nodes to reference, and a way to store them in config.
- Looks up the config values when building the block, rather than using taxonomy terms from the tags vocab.

Full disclaimer: relies on FCL nodes to exist to work. Migrations/shifting IDs may cause some headaches here unless we can keep block config up to date after each migration run.


DuttonMa (approved):
Nice work

PR #354, opened 2021-03-18
Translation help block

neilblair:
A simple block to output the translation help link plus javascript to process the Google Translate links on /articles/translation-help and add in the referring URL to be translated.

Related PRs: https://github.com/dof-dss/nidirect-drupal/pull/643, https://github.com/dof-dss/nicsdru_nidirect_theme/pull/258

### DIFF
--- a/nidirect_common/nidirect_common.module
+++ b/nidirect_common/nidirect_common.module
@@ -516,3 +516,20 @@ function nidirect_common_page_attachments(array &$page) {
 
   $page['#attached']['library'][] = 'nidirect_common/translation_help';
 }
+
+/**
+ * Implements hook_preprocess_node().
+ */
+function nidirect_common_preprocess_node(array &$variables) {
+
+  $node = $variables['node'];
+
+  if (isset($node) && method_exists($node, 'bundle')) {
+    // Fix for application pages to prevent Google Translate altering the green button link URL.
+    // (We don't want applications opening in Google Translate).
+    if ($node->bundle() === 'application') {
+      $variables['#attached']['library'][] = 'nidirect_common/application_links';
+    }
+  }
+
+}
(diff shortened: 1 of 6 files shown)


DuttonMa (changes requested):
This solution works well. My only slight reservation is in the hard coding of the node id 13488.
Could you replace this by looking for the alias and deriving the node id that way?

\Drupal::service('path.alias_manager')->getPathByAlias('/articles/translation-help')

nidirect_common/src/Plugin/Block/TranslationHelpBlock.php
```diff
@@ -0,0 +1,117 @@
...
+    $block_content = [
+      '#attributes' => [
+        'class' => ['section-translation-help'],
+      ]
+    ];
+
+    if ($translation_help_url) {
+      // Get the currently requested uri and add as query param to translation help url.
+      $uri = $this->requestStack->getCurrentRequest()->getRequestUri();
+      $translation_help_url->setOption('query', ['uri' => $uri]);
```
DuttonMa, line 86: I like this approach !

nidirect_common/nidirect_common.module
```diff
@@ -506,3 +506,34 @@ function nidirect_common_views_pre_view(ViewExecutable $view) {
...
 }
+
+/**
+ * Implements hook_page_attachments().
+ */
+function nidirect_common_page_attachments(array &$page) {
+
+  // Add translation help library to the translation help page.
+
+  if (\Drupal::service('path.current')->getPath() !== '/node/13488') {
```
DuttonMa: Could we get rid of the hard coding here as well?
Should just be a case of one call to \Drupal::service('path.alias_manager')->getPathByAlias('/articles/translation-help')

DuttonMa (approved):
All looks great to me

PR #367, opened 2021-04-20
D8 nid 1128

omahm:


nidirect_common/inc/preprocess.inc
```diff
@@ -20,13 +20,15 @@ function nidirect_common_preprocess_views_view(&$variables) {
...
+
   if (\Drupal::service('router.admin_context')->isAdminRoute()) {
+    $variables['#attached']['library'][] = 'nidirect_common/admin_theme';
     return;
   }
 
   // Similar to how core adds the node object to $variables, we will add
   // the term if we are on a taxonomy page.
-  if (\Drupal::routeMatch()->getRouteName() == 'entity.taxonomy_term.canonical') {
+  if ($route_name == 'entity.taxonomy_term.canonical') {
```
DuttonMa, line 31: Won't this fail ? How does the $route_name variable get set ?

DuttonMa (changes requested):

nidirect_common/inc/preprocess.inc
omahm: Quite right, I had that assigned locally but it's not in the branch..  ...adding it back in

DuttonMa (approved):

PR #369, opened 2021-04-25
Landing pages changes

neilblair:
Hide title for video blocks and plain card decks.  
Change field_teaser in blocks to use a textarea rather than single line input.


omahm (approved):
lg

PR #374, opened 2021-04-28
Add process callback on element format info

johangant:
Allows us to restrict editor options on layout builder block routes


omahm (approved):
Looks good

PR #377, opened 2021-05-06
Adjustment to contacts sorting process

johangant:
Slightly convoluted extra preprocesing needed to take into account case sensitivity of array_multisort and the odd thing views needs with the 'index' property which we can't update in array_multisort at the same time.


omahm (approved):
Nice one!!

PR #392, opened 2021-06-03
Adjust the telephone text for international freephone numbers.

johangant:
Underlying library struggles to parse these correctly and adding a fix there is significantly more complex.


omahm (approved):
Yup, the libphonenumber can be a pain to work with

PR #394, opened 2021-06-04
Disable form actions if the node has the locked flag set.

johangant:
Relates to https://github.com/dof-dss/nidirect-drupal/pull/722


omahm (approved):
Nice one

PR #395, opened 2021-06-04
D8NID-1253 flagging handler

johangant:
A bit rough around the edges but it seems to now fix the correct flag value for these so editors can't edit them once imported. Doesn't seem to clash with the general migrate post handler as it checks for the migration id to come from MAS import migration.


nidirect_money_advice_articles/src/EventSubscriber/PostMigrationSubscriber.php
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Set locked content flag for MAS articles.
+   *
+   * @return string
+   *   Info about the result.
+   * @throws \Exception
+   */
+  protected function processFlags() {
+    // Clear out the flag_counts table so we can reinsert new values.
```
omahm, line 88: This makes me nervous, what of there are articles locked but not part of the MAS, they'll be unlocked.

omahm (approved):
Approving but with 1 concern.

PR #396, opened 2021-06-07
Improved flag data deletion

johangant:
Now safer by using an inner join on MAS migration map table to ensure only MAS articles locked flag data is purged before re-importing. Means other non-MAS nodes can have locked flags kept without being affected by MAS imports.


omahm (approved):
Nice one

PR #401, opened 2021-06-09
Give a regional hint to geocoding if the address string is detected as having a UK postcode present

johangant:
Works nicely, once I detected the unadvertised hook in the geocoder module :)

`$this->moduleHandler->alter('geocode_address_string', $address_string);`


omahm (approved):
Noice!

PR #402, opened 2021-06-10
Remove duplicates from some theme listings

johangant:
De-dupe content render array by using the entity id as a key when adding to the collection to be rendered. The ordering isn't an issue as the array is sorted before being handed over to the theme layer.


omahm (approved):
LG

PR #406, opened 2021-06-21
D8NID-1214 ~ Fix prefix issue when identical numbers are used

omahm:
Ain't gonna win no beauty contest but fixes the issue when the contact uses the same number for text phone and other contact labels.


johangant (approved):
If it works, it works 😄

PR #410, opened 2021-06-24
D8NID-1253

johangant:
Super irritating process plugin order needed otherwise body text isn't saved/silently fails. See https://www.drupal.org/docs/8/api/migrate-api/migrate-process-plugins/migrate-process-overview for detail


DuttonMa:
Good find ! That certainly is annoying - we must remember it for other sites

DuttonMa (approved):

PR #418, opened 2021-07-07
Validation callback for specific webforms

johangant:
Prevents submission of HTML; supports the site without the use of honeypot to block unhelpful form submissions


omahm (approved):
Interesting way of detecting HTML 👍

PR #422, opened 2021-07-14
D8 nid 1363

DuttonMa:
Create new nidirect_dblog module and move hook_entity_delete processing over from nidirect_common.

Implement nidirect_dblog_cron to override dblog_cron - the new processing is more selective and does not delete certain log entries that we would like to keep (redirect deletion and content updates).

See associated PR https://github.com/dof-dss/nidirect-drupal/pull/780


johangant (approved):
Looks good, nice one. In the longer term, I think it'd be interesting to keep a separate audit trail for certain site tasks and operations... but definitely another conversation!

PR #423, opened 2021-07-16
Set webform node cache context to only vary by URL path

johangant:
The forms only vary by URL, this should cut down on extra cache items for other contexts that aren't expected to give any variance in render ouput.


DuttonMa (approved):
Looks good

PR #448, opened 2021-11-03
D8 nid 1388

johangant:
Without this, changing a LP's theme value will leave a dangling redirect which can be pruned away manually easily enough but is causing confusion even for seasoned NIDirect testers, let alone casual editors.

Adding a taxo list to expire too so when the change is made, the taxo list for the theme previously used is also expired as well as a parent term list which landing pages also seem to need.


omahm (approved):
LGTM

PR #465, opened 2022-05-26
Event subscriber to change cache header on ED waiting times page

neilblair:
This borrows code Johan wrote to prevent Fastly caching the school closures page.  Like the school closures page, the ED waiting times page uses a token to output data from a feed.  The feed is updated every hour, but since the entire page is cached by Fastly for 24 hours (due to the surrogate-control max-age header set on all pages) then updated data never appears on the page when it should.  So using an event subscriber, we detect if the response contains rendered hospital waiting times and then change the surrogate-control max-age header to a much shorter period.


omahm (approved):
LG

PR #471, opened 2022-06-07
Hide the revert to default layout button if non-admin user. (#469)

johangant:
Prevents layout data loss on content with multiple revisions


johangant:
Closing off, for now, until QA complete + wider PR created for multiple issues to be released.

PR #474, opened 2022-06-22
Do not output school closures on a specific path

neilblair:
(already output via token replacement)


omahm (changes requested):
Looks good although with nidirect_school_closures_render() only called from one place I'd be tempted to remove that function completely and drop the render array creation from it straight into the token output.

omahm (approved):
👍 LG

PR #478, opened 2022-07-22
Update to PHPstan level 2 compliance

johangant:
TL;DR:

* Many classes have property namespaces missing a `\` prefix. NB: not required in the `Use` statements.
* Lots of type hinting now in place to guide the IDE/static analysis tool (phpstan) to assess what functions/properties are available for use.
* Reduction in magic methods (`$node->field_body->value`) which, while functional (for now) make it impossible to see the contents until runtime.
* Two cases of ignoring phpstan, because I can't work out what type the next thing should be considered even from the class reflection/inspection. :man_shrugging:


neilblair (approved):
Oof 62 files changed! LGTM

PR #480, opened 2022-08-31
D8NID-1540: convert relative URLs to absolute URLs in services catalogue json feed

neilblair:
Custom serializer for view generating services-catalogue.json to ensure relative URLs in content are converted to absolute URLs.


omahm (approved):
LG

johangant:
Looks good to me. You'll get the deprecated code task failing because it's unhappy about the cookie notice module's patch that needs taking out of the composer file. I think that's approved in another PR so once the two balance out you should be fine to proceed.

PR #484, opened 2022-10-13
Log a warning if missing breadcrumb is detected on a node

neilblair:
The mystery of the disappearing breadcrumb trail continues.  Had another incident yesterday where a user reported it was missing.  I tried downloading the db and importing locally - but the trail worked ok. 

Hoping this hook_preprocess_breadcrumb will log an error as soon as breadcrumbs start to disappear on live nodes. I can get logz.io to send an email alert if a log entry about missing breadcrumbs appears ... That's the theory anyway!

Can maybe expand this with additional code to troubleshoot why the breadcrumb is missing?

### DIFF
--- a/nidirect_breadcrumbs/nidirect_breadcrumbs.module
+++ b/nidirect_breadcrumbs/nidirect_breadcrumbs.module
@@ -52,10 +52,8 @@ function nidirect_breadcrumbs_preprocess_breadcrumb(&$variables) {
 
   if (\Drupal::state()->get('missing_breadcrumb_check') !== 1) {
     $route_match = \Drupal::routeMatch();
-    // Log a warning if a node has no breadcrumb trail.
     if ($route_match->getRouteName() == 'entity.node.canonical') {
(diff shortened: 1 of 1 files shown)


johangant (commented):
Interesting idea. In terms of things that might be useful to know - in addition to the page on which it might happen - things like a dump of the $variables collection as close as possible to the point where they're about to be rendered. It's a big array so could be serialised to a php string. It might be very large though, so if logz.io struggles or has a limit something like string compression or encoding could help?

neilblair:
> Interesting idea. In terms of things that might be useful to know - in addition to the page on which it might happen - things like a dump of the $variables collection as close as possible to the point where they're about to be rendered. It's a big array so could be serialised to a php string. It might be very large though, so if logz.io struggles or has a limit something like string compression or encoding could help?

👍🏻 I think logz.io is pretty generous on log line length - 500,000 bytes. Hopefully that's enough

johangant (approved):

PR #501, opened 2023-09-12
Prison visits

neilblair:
Updated config for prison visits form plus changes to the webformhandler to add alterForm method which filters time slots shown to the user based on values extracted from the booking reference number and values from config.


omahm (approved):
Nice one, looks really good.

PR #502, opened 2023-09-15
Prison visits

neilblair:
Additional logic for webformhandler when filtering time slots:
- Time slots available take account of advance notice required for booking different types of visit.
- "Separates" get either AM or PM time slots (alternating from week to week) for certain visit types and prisons


johangant (approved):
LGTM :+1:

PR #504, opened 2023-09-21
Prison visits

neilblair:
Beefier JS validation for prison visit booking reference to check there is a bookable time slot before the expiry date of booking reference.  Its complicated because the last available time slot alternates between AM and PM from week to week for "separates" prisoners and a time slot must be booked 24 hours in advance for face-to-face visits and 48 hours in advance for virtual visits ...


omahm (approved):
That's some work 👍

